PR #148, opened 2020-02-06
CGPROD 2134 Safe Area Min Width

FostUK:
* Split safeArea into own file.
* topLeft and topRight groups are set to fixed width of 64.
* gets nearest of array of buttons to create safe area.


test/core/layout/safe-area.test.js
```diff
@@ -0,0 +1,77 @@
...
+ * @copyright BBC 2020
+ * @author BBC Children's D+E
+ * @license Apache-2.0
+ */
+
+/**
+ * @copyright BBC 2018
+ * @author BBC Children's D+E
+ * @license Apache-2.0
+ */
```
AdamXA: two of these

FostUK: Fixed ta!

src/core/layout/layout.js
```diff
@@ -150,7 +128,7 @@ export function create(scene, metrics, buttonIds) {
         addToGroup,
         buttons,
         debug,
-        getSafeArea,
+        getSafeArea: getSafeArea(groups),
```
AdamXA: Just to clarify, this will return a func for the original GelGroups? Any groups added after, are not used I guess, guess thats fine

FostUK: Yeah, but I think custom groups would be unlikely to be part oif the calculation.

On thought -  do you think it would be clearer if I called the function in safe-area-js "getSafeAreaFn" ?

so that line would look like: getSafeArea: getSafeAreaFn(groups),

Or does that make it worse?

AdamXA: I guess it may be a bit clearer that its returning a function, yea

FostUK: Cool - have updated, at least illustrates that the `getSafeArea` hanging off layout is a different thing to the one imported from safe-area.js

PR #153, opened 2020-02-07
CGPROD-2121 Grid cell alignment fix

AnthonyGoodchild:
https://jira.dev.bbc.co.uk/browse/CGPROD-2121

Using an alignment formula to calculate the X/Y position of the cells, based on the cell's column and the total number of cells per row.

This new formula brings the cells together, so that the spacing between them (padding) is consistently in line with the gel specification.

Screen Shot 2020-02-07 at 17 42 34


src/core/layout/grid/cell.js
```diff
@@ -37,35 +37,34 @@ export const setSize = (grid, button) => {
...
+    button.y = button.displayHeight * alignFactorY + grid._cellPadding * alignFactorY;
 };
 
 export const create = (grid, choice, idx) => {
     const config = {
         ...choice,
         scene: grid.scene.scene.key,
         channel: grid.eventChannel,
-        tabbable: true,
+        alwaysTab: grid.cellsPerPage === 1,
```
FostUK: This line should be set to how it is on master - needed by Jacob's work.

PR #154, opened 2020-02-10
[CGPROD-2132] Countup audio changes for low and high scores

AdamXA:
Update counters using requestAnimationFrame -> update loop
This should prevent count ups taking longer than they should when ticks per second is higher than the game frame rate.

Small scores just play audio for each tick - define this range in singleTicksRange
For higher scores, define a ticksPerSecond - the rate at which audio plays above this range.


AnthonyGoodchild:
〰️ 🔫

PR #163, opened 2020-02-11
CGPROD 2137 IE11 build issue

FostUK:
IE11

Removes Proxy from crel as babel has no way to reliably polyfill it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,6 @@
         "babel-jest": "^24.8.0",
         "babel-loader": "^8.0.6",
         "babel-plugin-module-resolver": "^3.2.0",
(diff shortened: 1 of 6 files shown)


AnthonyGoodchild:
I don't like this whole thing but ok

package.json
```diff
@@ -20,8 +20,8 @@
         "babel-jest": "^24.8.0",
         "babel-loader": "^8.0.6",
         "babel-plugin-module-resolver": "^3.2.0",
+        "babel-plugin-proxy": "^1.1.0",
```
AdamXA: not used right?

FostUK: Ah yeah. Will remove.

PR #167, opened 2020-02-12
Cgprod 2005 results accessibility

alex-s2:
Allow screen reader to read scores.


src/core/accessibility/accessibilify.js
```diff
@@ -43,28 +43,42 @@ const assignEvents = (accessibleElement, button) => {
...
-        onMouseOut: mouseOut,
-    };
+    if (interactive === true) {
+        const buttonAction = () => {
+            if (!button.input.enabled) return;
+            button.emit(Phaser.Input.Events.POINTER_UP, button, sys.input.activePointer, false);
+        };
+        const mouseOver = () => button.emit(Phaser.Input.Events.POINTER_OVER, button, sys.input.activePointer, false);
+        const mouseOut = () => button.emit(Phaser.Input.Events.POINTER_OUT, button, sys.input.activePointer, false);
+        options = {
```
JaKraken: Could we pull this out this duplicated definition of options and use a ternary? something like:
```
const options = {
            id,
            button,
            class: "gel-button",
            "aria-label": button.config.ariaLabel,
            onClick: onClick ? onClick : () => {},
            onMouseOver:  onMouseOver ? onMouseOver : () => {},
            onMouseOut: onMouseOut ? onMouseOut : () => {},
            interactive: interactive,
            parent: interactive ? sys.scale.parent : undefined,
        };
```

src/core/accessibility/accessible-dom-element.js
```diff
@@ -15,6 +15,11 @@ const keyUp = options => event => {
     }
 };
 
+let defaultAttributes = {
```
JaKraken: Can use const here since you're only reasigning a property of the object

src/core/accessibility/accessibilify.js
```diff
@@ -43,28 +43,42 @@ const assignEvents = (accessibleElement, button) => {
     resizeAndRepositionElement();
 };
 
-export function accessibilify(button, gameButton = true) {
+export function accessibilify(button, gameButton = true, interactive = true) {
     const sys = button.scene.sys;
     const scene = button.scene;
     const id = [scene.scene.key, button.config.id].join("__");
+    var options;
```
AdamXA: var

src/components/results/results-row.js
```diff
@@ -45,6 +52,14 @@ export class ResultsRow extends Phaser.GameObjects.Container {
 
         this.rowConfig.format &&
             this.rowConfig.format.forEach(object => {
+                if (object.type === "text") {
```
AdamXA: maybe `||` it?

alex-s2: they do different stuff

src/core/accessibility/accessibilify.js
FostUK: Yeah, or create a defaults object with the on mouseover partsand just do 

options = { ...defaults, ...newStuff}

?

PR #168, opened 2020-02-12
[CGPROD-2152] Added a script to copy starter pack docs and theme config

AdamXA:
Will get ran by the CI - genie-starter-pack-copy-job


Leonie-:
Should this be in the Jenkins execute shell section rather than in the repo?

AdamXA:
Decided to move script out of source control and into jenkins

PR #174, opened 2020-02-12
Fix for cells shifting to incorrect positions

AnthonyGoodchild:
when tweening on pagination the page blank cell count was being
calculated from the `grid.page` value, which is not consistently the
correct page index that the cell belongs to.

This change calculates the `page` value in the calling method, and passes
in to the blank cell calculation, therefore always returns the correct
blank cell count for the page it belongs to.


src/core/layout/grid/cell.js
```diff
@@ -40,18 +40,19 @@ export const setSize = (grid, button) => {
...
 
 const setPosition = (grid, button, idx) => {
     const pageIdx = idx % grid.cellsPerPage;
+    const page = Math.floor(idx / grid.cellsPerPage);
     const col = pageIdx % grid._config.columns;
     const row = Math.floor(pageIdx / grid._config.columns);
 
     const alignFactorX =
-        col + getBlankCellCount(grid, row) * alignmentFactor[grid._config.align] - (grid._config.columns - 1) / 2;
+        col + getBlankCellCount(grid, row, page) * alignmentFactor[grid._config.align] - (grid._config.columns - 1) / 2;
```
FostUK: Is there any test we can stick round this?

AnthonyGoodchild: no

PR #176, opened 2020-02-13
Cgprod 2112 single item pause

FostUK:
* Remove all events created by single item mode on shutdown
* Call shutdown on overlay created
* Call single item mode on overlay closed

### DIFF
--- a/src/components/select/select-screen.js
+++ b/src/components/select/select-screen.js
@@ -20,12 +20,11 @@ const gridDefaults = {
 };
 
 const getOnTransitionStartFn = scene => () => {
-    if (!scene.layout.buttons.continue || !scene.layout.buttons.continue.accessibleElement) return;
+    if (!scene.layout.buttons.continue) return;
 
     const bool = scene.currentEnabled();
     scene.layout.buttons.continue.input.enabled = bool;
     scene.layout.buttons.continue.alpha = bool ? 1 : 0.5;
-    scene.layout.buttons.continue.accessibleElement.update();
 };
 
 export class Select extends Screen {
--- a/src/components/select/select-screen.js
+++ b/src/components/select/select-screen.js
@@ -32,7 +32,7 @@ export class Select extends Screen {
     create() {
         this.add.image(0, 0, `${this.scene.key}.background`);
         this.addAnimations();
-        this.theme = this.context.config.theme[this.scene.key];
+        this.theme = this.context.theme;
         this.titles = createTitles(this);
         const buttons = ["home", "pause", "previous", "next"];
         this.setLayout(buttons.concat(singleItemMode.continueBtn(this)));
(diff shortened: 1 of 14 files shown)


src/components/select/select-screen.js
```diff
@@ -86,6 +86,16 @@ export class Select extends Screen {
...
 
+    addOverlay(key) {
+        this.singleItemMode.shutdown();
+        super.addOverlay(key);
+    }
+
+    _onOverlayRemoved(overlay) {
+        super._onOverlayRemoved(overlay);
+        this.singleItemMode = singleItemMode.create(this);
+    }
```
AdamXA: Shouldn't need this stuff if we use makeAccessible - I'll have a look at a way to do this today

test/components/select/select-screen.test.js
```diff
@@ -316,6 +342,46 @@ describe("Select Screen", () => {
         });
     });
 
+    describe("_onOverlayRemoved method", () => {
+        test("calls super method with same args", () => {
+            selectScreen._onOverlayRemoved("testOverlay");
+
+            expect(Screen.mock.instances[0]._onOverlayRemoved).toHaveBeenCalledWith("testOverlay");
+        });
```
AdamXA: this test passes without the _onOverlayRemoved function in the select screen.

test/components/select/select-screen.test.js
```diff
@@ -316,6 +342,46 @@ describe("Select Screen", () => {
...
+                bringToTop: jest.fn(),
+            };
+
+            selectScreen.singleItemMode = {
+                shutdown: jest.fn(),
+            };
+            selectScreen.addOverlay("testKey");
+
+            expect(Screen.mock.instances[0].addOverlay).toHaveBeenCalledWith("testKey");
+        });
```
AdamXA: so does this one

src/components/select/single-item-mode.js
```diff
@@ -6,46 +6,46 @@
...
 export const create = scene => {
     const continueButton = scene.layout.buttons.continue;
     if (!continueButton) return false;
 
-    a11y.removeButton(continueButton);
-    a11y.removeButton(scene.layout.buttons.next);
-    a11y.removeButton(scene.layout.buttons.previous);
-    a11y.reset();
+    const onBlur = () => scene.grid.showPage(0);
+    window.addEventListener("blur", onBlur);
```
AnthonyGoodchild: This is causing undesired behaviour when switching browser tab -> switching back.
It looks like it jumps to the first cell then animates to the second without showing the asset.

Could it be related to the addition to the return in `showPage` ?
https://github.com/bbc/genie/pull/176/files#diff-b5d608f4ffc03234f3a1a5250497a26eR87

AdamXA: can reproduce, will try and find out what it is now

PR #183, opened 2020-02-18
[CGPROD-2200] Fix buttons not being tabbable after resets

AdamXA:
Not sure why we check if a button is focused and hide/disable it. But it's causing a bug that intermittently occurs.


AnthonyGoodchild:
😂 😂 😂

AdamXA:
Been through it and got the verbal approval 👍

PR #184, opened 2020-02-19
CGPROD-2206_Particle-Backgrounds-for-Results-Screen 

Leonie-:
Adds particle backgrounds to results screen. May also want to devise some way of having Phaser Geom objects defined in the config.


AdamXA:
Closing this as background particles have been implemented (I think) :D

PR #192, opened 2020-02-24
CGPROD-2231: Undefined Stat for Single Item Select Screen

JaKraken:
Clicking continue does not pass a callback returning the same object as when clicking the gel button, this meant that selection.title was undefined when pressing continue. 

To fix this we now pass the currently selected button's title and id via a callback into the next function on select screen.

### DIFF
--- a/src/core/layout/grid/grid.js
+++ b/src/core/layout/grid/grid.js
@@ -66,6 +66,10 @@ export class GelGrid extends Phaser.GameObjects.Container {
         });
     }
 
+    getCurrentSelection() {
+        return this.choices()[this.page];
+    }
+
     enforceLimits() {
         const maxColumns = this._config.rows === 1 ? 4 : 3;
         const maxRows = 2;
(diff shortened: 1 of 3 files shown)


src/core/layout/grid/grid.js
```diff
@@ -66,6 +66,10 @@ export class GelGrid extends Phaser.GameObjects.Container {
         });
     }
 
+    getCurrentSelection() {
+        return this.choices()[this.page];
```
AnthonyGoodchild: I hate this but it's just how we are doing it!

Leonie-: How would you want to do it instead?

Leonie-: A getter?

AnthonyGoodchild: I just don't like `this.page` === selection. I wouldn't change it yet as we don't really require any more to be done in this area - and this works!

PR #193, opened 2020-02-24
CGPROD 2067 Carousel tabs

FostUK:
Disables accessible elements for previous and next when they are hidden


test/components/how-to-play.test.js
```diff
@@ -75,7 +74,13 @@ describe("How To Play Screen", () => {
         howToPlayScreen.scene = { key: "test-select" };
         howToPlayScreen.game = { canvas: { parentElement: "parent-element" } };
         howToPlayScreen.navigation = { next: jest.fn() };
+        howToPlayScreen.sys = {
+            queueDepthSort: jest.fn(),
+        };
         howToPlayScreen.setLayout = jest.fn(() => mockLayout);
+        Object.defineProperty(howToPlayScreen, "layout", {
+            get: jest.fn(() => mockLayout),
```
Leonie-: Not sure why this mock is needed? Will have a look...

Leonie-: Seems like it isn't?

Leonie-: Sorry - that should be line 77-79

test/mock/dom-element.js
```diff
@@ -29,6 +29,7 @@ const domElement = () => {
         addEventListener: jest.fn(),
         removeEventListener: jest.fn(),
         childNodes: [],
+        update: jest.fn(),
```
Leonie-: This is not needed either...

src/components/how-to-play.js
```diff
@@ -29,9 +29,9 @@ export class HowToPlay extends Screen {
             : this.add.image(0, -170, `${this.scene.key}.title`);
 
         if (this.theme.howToPlay) {
-            this.buttonLayout = this.setLayout(["overlayBack", "audio", "settings", "previous", "next"]);
+            this.setLayout(["overlayBack", "audio", "settings", "previous", "next"]);
         } else {
-            this.buttonLayout = this.setLayout(["home", "audio", "pause", "previous", "next", "continue"]);
+            this.setLayout(["home", "audio", "pause", "previous", "next", "continue"]);
```
Leonie-: Is this logic fork still used? Isn't it always How To Play now?

src/components/how-to-play.js
```diff
@@ -47,14 +47,17 @@ export class HowToPlay extends Screen {
...
 
     setButtonVisibility() {
-        this.buttonLayout.buttons.previous.visible = Boolean(!this.theme.howToPlay || this.currentIndex !== 0);
+        this.layout.buttons.previous.visible = Boolean(!this.theme.howToPlay || this.currentIndex !== 0);
 
         const isNotLastPage = this.currentIndex + 1 !== this.choiceSprites.length;
-        this.buttonLayout.buttons.next.visible = Boolean(!this.theme.howToPlay || isNotLastPage);
+        this.layout.buttons.next.visible = Boolean(!this.theme.howToPlay || isNotLastPage);
+
+        this.layout.buttons.previous.accessibleElement.update();
```
Leonie-: Do we need tests for these new lines?

test/components/how-to-play.test.js
FostUK: ah! Good spot. Was a fix for an intermediate issue I left behind. Updated :)

src/components/how-to-play.js
FostUK: That's a good point. Looks like Alexa added it in originally - will double check with her when she gets in.

src/components/how-to-play.js
FostUK: Visibility is tested in the fixed tests below.
Testing update is called feels like the wrong thing to check. Will have a think...

src/components/how-to-play.js
FostUK: Yeah uneeded due to new select screen :)

src/components/how-to-play.js
FostUK: Ended up testing update is called.
Not sure that's the best thing but it's mocked so prob all that can be done.

PR #205, opened 2020-03-17
[NOTICKET] Bump dependency versions

AdamXA:
Doesn't fix security vulnerability, but doesn't seem to break anything either.

Edit: apart from the license checker 💤


AdamXA:
danke schön

PR #209, opened 2020-04-01
Noticket global scripts

FostUK:
11_nights_b_1

* Force load order with dynamic scripts in both local dev and final build paths


Leonie-:
Looks good, but no tests? :maple_leaf:

FostUK:
> Looks good, but no tests? 🍁
Yeah - back when these were originally added we found it impossible to write a test around webpack functionality so decided dev / build scripts didn't need them

Possibly worth having another go now in the Jest world but think will do that in a separate ticket so we can get this fix to Aerian.

PR #210, opened 2020-04-03
NOTICKET particles example page

FostUK:
hana-bi

* Adds particles plugin and background loader example
* Renames background-animations system to background furniture
* Inadvertently implements CGPROD-2206 (background particles)
* Adds basic particle loader plugin (does not support geometry yet)


debug/examples/example-files.json
```diff
@@ -26,6 +26,11 @@
                 "type": "json5",
                 "key": "background-animations",
                 "url": "background-animations.json5"
+            },
+            {
+                "type": "json5",
+                "key": "background-particles",
```
AdamXA, line 32: Is this going to be loaded as a particles file in future?

AdamXA: Oh wait sorry, this is a screen isn’t it 😂

FostUK: Ha! 
Yeah I actually debated renaming those to furniture but I think they make more sense as is.

PR #217, opened 2020-04-14
NOTICKET Select Fix

FostUK:
* Select screen uses configured storageKey instead of "levels" hardcoded.

### DIFF
--- a/test/components/select/titles.test.js
+++ b/test/components/select/titles.test.js
@@ -15,22 +15,24 @@ describe("Select Screen - Titles", () => {
         mockText = { setOrigin: jest.fn() };
         mockScene = {
             scene: { key: "character-select" },
-            theme: {
-                titles: [
-                    { type: "image", key: "title", xOffset: 0, yOffset: -250 },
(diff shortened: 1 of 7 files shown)


test/components/select/titles.test.js
```diff
@@ -15,22 +15,24 @@ describe("Select Screen - Titles", () => {
...
-                    { type: "text", value: "Select Your Character", xOffset: 0, yOffset: -260 },
-                    { type: "text", value: "Sub-Title", xOffset: 0, yOffset: -233 },
-                ],
-            },
             add: { image: jest.fn(() => mockImage), text: jest.fn(() => mockText) },
             assetPrefix: "character-select",
+            context: {
+                theme: {
+                    titles: [
+                        { type: "image", key: "title", xOffset: 0, yOffset: -250 },
```
AnthonyGoodchild, line 23: should this be called `offsetX` and `offsetY` instead?

FostUK: Yes. This is a prime example of what I was talking about. x and y would be better.

AnthonyGoodchild: 🤦

PR #218, opened 2020-04-14
[CGPROD-2204] Parse JSON into Phaser.Geoms

AdamXA:
just the geom parsing stuff in this PR

### DIFF
--- a/src/core/loader/particles-loader/geom-parse.js
+++ b/src/core/loader/particles-loader/geom-parse.js
@@ -1,0 +1,16 @@
+/**
+ * @copyright BBC 2020
+ * @author BBC Children's D+E
+ * @license Apache-2.0
+ */
+const geomObjects = {
+    rectangle: config => new Phaser.Geom.Rectangle(config.x, config.y, config.width, config.height),
+    circle: config => new Phaser.Geom.Circle(config.x, config.y, config.radius),
+    ellipse: config => new Phaser.Geom.Ellipse(config.x, config.y, config.width, config.height),
+    line: config => new Phaser.Geom.Line(config.x1, config.y1, config.x2, config.y2),
+    point: config => new Phaser.Geom.Point(config.x, config.y),
+    polygon: config => new Phaser.Geom.Polygon(config.points),
+    triangle: config => new Phaser.Geom.Triangle(config.x1, config.y1, config.x2, config.y2, config.x3, config.y3),
+};
+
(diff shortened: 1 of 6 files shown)


src/core/loader/particles-loader/geom-parse.js
```diff
@@ -0,0 +1,16 @@
...
+    rectangle: config => new Phaser.Geom.Rectangle(config.x, config.y, config.width, config.height),
+    circle: config => new Phaser.Geom.Circle(config.x, config.y, config.radius),
+    ellipse: config => new Phaser.Geom.Ellipse(config.x, config.y, config.width, config.height),
+    line: config => new Phaser.Geom.Line(config.x1, config.y1, config.x2, config.y2),
+    point: config => new Phaser.Geom.Point(config.x, config.y),
+    polygon: config => new Phaser.Geom.Polygon(config.points),
+    triangle: config => new Phaser.Geom.Triangle(config.x1, config.y1, config.x2, config.y2, config.x3, config.y3),
+};
+
+export const geomParse = geomConfig => geomObjects[geomConfig.type](geomConfig);
```
FostUK, line 16: I like this line.

AdamXA: what in particular do you like about this line

test/core/loader/particles-loader/particle-file.test.js
```diff
@@ -39,6 +42,54 @@ describe("Particles File", () => {
...
+            geomParse.mockImplementation(() => "mock emit zone");
+            file.xhrLoader = {
+                responseText: testJSON,
+            };
+            file.onProcessComplete = jest.fn();
+            file.onProcess();
+            expect(file.data).toEqual({ emitZone: { source: "mock emit zone" } });
+        });
+
+        test("calls the the Geom parser when a deathZone source is provided", () => {
```
Leonie-, line 69: Do we need a test to check that the Geom parser is not called when there is niether a death zone or an emit zone?

src/core/loader/particles-loader/geom-parse.js
Leonie-: It is very terse, and works nicely with the object literal above.

test/core/loader/particles-loader/particle-file.test.js
AdamXA: added, thanks 👍

PR #224, opened 2020-04-16
[CGPROD-2204] Fix particle emitter being covered by row backdrop

AdamXA:
- Moves backdrop out of the container so that particle emitters can be placed between the row backdrop and the container more easily - use of containers made this trickier.
- Fixes bug when an alpha of 0 set on central backdrop, would set the alpha to 1.
- Since was messing around with backdrops and scaling, the results screen should now properly scale so that it looks the same at all resolutions.
- Added row particles example to genie example launcher and fixed existing examples.


src/components/results/results-row.js
```diff
@@ -76,19 +75,6 @@ export class ResultsRow extends Phaser.GameObjects.Container {
...
     }
 
-    createBackdrop() {
-        fp.get("backdrop.key", this.rowConfig) && this.backdropFill();
-    }
-
-    backdropFill() {
-        const config = this.rowConfig.backdrop;
-        const backdrop = this.scene.add.image(config.offsetX || 0, config.offsetY || 0, config.key);
-        backdrop.alpha = config.alpha || 1;
```
AdamXA, line 86: eg. 0 || 1 = 1

not 0 😀

src/components/results/results-screen.js
```diff
@@ -38,31 +42,36 @@ export class Results extends Screen {
...
+        tweenRowBackdrops(this, this.rowBackdrops, this.rows.containers);
         playRowAudio(this, this.rows.containers);
         addParticlesToRows(this, this.rows.containers);
     }
 
-    createBackdrop() {
-        fp.get("backdrop.key", this.context.theme) && this.backdropFill();
-        this.sizeToParent(this.backdrop, this.resultsArea());
+    createCentralBackdrop() {
+        fp.get("backdrop.key", this.context.theme) && this.centralBackdropFill();
```
FostUK, line 53: One day... optional chaining will solves all this...

PR #227, opened 2020-04-21
[CGPROD-2204] Results row emitter above content

AdamXA:


themes/default/config/results.json5
```diff
@@ -105,6 +105,7 @@
                             offsetY: -15,
                             delay: 1000,
                             duration: 1000,
+                            placeAboveContent: false,
```
FostUK, line 108: could this just be "placeAbove" or even just "onTop" ?

Worried lately that configs are starting to become too complicated to pick up easily for agencies. We could do with a simplification pass on them to see if there's anything we can do about it..

AdamXA: sure

PR #228, opened 2020-04-21
[CGPROD-2321] Add particles on hovering select cells

AdamXA:


src/components/select/select-particles.js
```diff
@@ -0,0 +1,27 @@
+/**
+ * @copyright BBC 2020
+ * @author BBC Children's D+E
+ * @license Apache-2.0
+ */
+import { gmi } from "../../core/gmi/gmi.js";
+
+const onPointerOut = (emitter, button) => button.on(Phaser.Input.Events.POINTER_OUT, () => emitter.stop());
```
FostUK, line 8: button.on(Phaser.Input.Events.POINTER_OUT, () => emitter.stop());
does 
button.on(Phaser.Input.Events.POINTER_OUT, emitter.stop);

work? might get binding errors...

src/components/select/select-particles.js
```diff
@@ -0,0 +1,27 @@
...
+const createEmitter = (scene, button, config) => {
+    const emitter = scene.add
+        .particles(config.assetKey)
+        .createEmitter(scene.cache.json.get(config.emitterConfigKey))
+        .setPosition(button.x, button.y);
+    onPointerOut(emitter, button);
+};
+
+export const addHoverParticlesToCells = (scene, cells, config, layoutRoot) => {
+    if (!config) {
```
FostUK, line 19: Can do if (!config) return; 

if you want

src/components/select/select-particles.js
AdamXA: na I get binding errors

src/components/select/select-particles.js
AdamXA: nice! 3 lines always felt too much for this! pretty sure I've done it elsewhere too

PR #230, opened 2020-04-27
Noticket game options

FostUK:
Splits phaser default config generation and adds ability to add basic game options from mian.js

### DIFF
--- a/src/core/loader/phaserDefaults.js
+++ b/src/core/loader/phaserDefaults.js
@@ -26,7 +26,7 @@ export const getPhaserDefaults = config => {
         ...{
             width: CANVAS_WIDTH,
             height: CANVAS_HEIGHT,
-            renderer: browser.forceCanvas ? Phaser.CANVAS : Phaser.AUTO,
+            type: browser.forceCanvas ? Phaser.CANVAS : Phaser.AUTO,
             antialias: true,
             multiTexture: true,
             parent: getContainerDiv(),
--- a/src/core/loader/phaserDefaults.js
+++ b/src/core/loader/phaserDefaults.js
@@ -1,3 +1,8 @@
+/**
+ * @copyright BBC 2020
+ * @author BBC Children's D+E
+ * @license Apache-2.0
+ */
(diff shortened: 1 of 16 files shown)


src/core/loader/phaserDefaults.js
```diff
@@ -0,0 +1,74 @@
...
+const getScenes = conf => Object.keys(conf).map(key => new conf[key].scene({ key, ...conf[key].settings }));
+
+export const getPhaserDefaults = config => {
+    const browser = getBrowser();
+    const scene = getScenes(Object.assign(config.screens, getLauncherScreen(debugMode.isDebug())));
+    scene.unshift(new Loader());
+    scene.unshift(new Boot(config.screens));
+
+    return {
+        ...{
```
AdamXA, line 26: does something different happen if you do `{ ...{ props }, ...otherprops }` or is this the same as`{ props, ...otherprops }`?

test/core/loader/container.test.js
```diff
@@ -0,0 +1,37 @@
+import { createMockGmi } from "../../mock/gmi.js";
+import { domElement } from "../../mock/dom-element.js";
+import { getContainerDiv } from "../../../src/core/loader/container.js";
+import { gmi } from "../../../src/core/gmi/gmi.js";
+
+/**
+ * @copyright BBC 2020
+ * @author BBC Children's D+E
+ * @license Apache-2.0
+ */
```
AdamXA, line 10: does it not complain that this is not at the top? 😲

FostUK: Ha. Will update. Good spot!

src/core/loader/phaserDefaults.js
FostUK: I've tended to do it by default because it's copying rather than mutating but in this case there's no point (since the object was just created anyway.

FostUK: Actually no what am I talking about. This is correct

{ props, ...otherprops }

will create a property called props with the object hanging off it.

AdamXA: think my example was a bit confusing 😂

PR #231, opened 2020-05-05
[CGPROD-2396] Enable pointerover/out events on disabled select items

AdamXA:
Allows particle effects to hook into the pointerover/out events.
Makes way for allowing separate particle emitter config for the items "disabled" state on pointerover/out.


src/components/select/select-screen.js
```diff
@@ -72,7 +72,7 @@ export class Select extends Screen {
             config.properties && Object.assign(button.sprite, config.properties);
 
             config.suffix && (button.config.ariaLabel = [button.config.ariaLabel, config.suffix].join(" "));
-            button.input.enabled = Boolean(config.enabled !== false);
```
AdamXA, line 75: wasn't tested originally so no test to remove

PR #234, opened 2020-05-06
[CGPROD-2388] Vertical border padding set to 32 on iPhone 5

AdamXA:


AdamXA:
> This is horrible.

It felt like I was committing a crime when I was writing it.

PR #241, opened 2020-05-14
Add level select button to pause screen if one is configured

alex-s2:
CGPROD-1937 Add level select button to pause screen if one is configured


src/components/overlays/pause.js
```diff
@@ -20,9 +20,20 @@ export class Pause extends Screen {
 
         const parentKey = this.context.parentScreens.slice(-1)[0].scene.key;
         const showReplay = Boolean(this.context.navigation[parentKey].routes.restart);
```
FostUK, line 22: Aha! I KNEW we'd done this before :)

alex-s2: Yeah in the same file too 😂

src/components/overlays/pause.js
```diff
@@ -20,9 +20,20 @@ export class Pause extends Screen {
...
 
-        this.setLayout(buttons.concat(replayButton));
+        if (showReplay && !showSelect) {
+            this.setLayout(buttons.concat(replayButton));
+        } else if (!showReplay && showSelect) {
+            this.setLayout(buttons.concat(selectButton));
+        } else if (showReplay && showSelect) {
+            this.setLayout(buttons.concat(replayAndSelectButton));
+        } else {
+            this.setLayout(buttons);
```
AdamXA, line 36: this could defo be tidied up a bit, something like:

```
showReplay && buttons.concat(replayButton)
showSelect && buttons.concat(selectButton)
```

alex-s2: You can't concat more than once it explodes

src/components/overlays/pause.js
```diff
@@ -20,9 +20,20 @@ export class Pause extends Screen {
 
         const parentKey = this.context.parentScreens.slice(-1)[0].scene.key;
         const showReplay = Boolean(this.context.navigation[parentKey].routes.restart);
+        const showSelect = Boolean(this.context.navigation["pause"].routes.select);
```
AdamXA, line 23: doesn't really matter, but you can just do `.navigation.pause.routes` instead of the `["pause"]` on these

src/components/overlays/pause.js
AdamXA: oh right you need to a setter then
```
showReplay && (buttons = buttons.concat(replayButton))
showSelect && (buttons = buttons.concat(selectButton))
```

alex-s2: I'm pretty sure it still doesn't work, there is something weird going on with buttons.concat

AdamXA: 🤔

alex-s2: Hm this specific way does work, was doing some other concating and it broke, changed thanks

PR #244, opened 2020-05-15
Fix results transientData from being overwritten

AnthonyGoodchild:


AdamXA:
Is this just so you can change it in debug before it goes to the next screen?

AnthonyGoodchild:
> Is this just so you can change it in debug before it goes to the next screen?

yeah

PR #245, opened 2020-05-18
NOTICKET background item names

FostUK:
* Adds names to background items for use by narrative animation system

### DIFF
--- a/src/core/background/furniture.js
+++ b/src/core/background/furniture.js
@@ -10,6 +10,11 @@ import { isSpine, addSpine } from "./spine.js";
 import { isText, addText } from "./text.js";
 import { isParticles, addParticles } from "./particles.js";
 
+const setName = createFn => config => {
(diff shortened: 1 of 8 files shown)


src/core/background/furniture.js
```diff
@@ -10,6 +10,11 @@ import { isSpine, addSpine } from "./spine.js";
 import { isText, addText } from "./text.js";
 import { isParticles, addParticles } from "./particles.js";
 
+const setName = createFn => config => {
+    const item = createFn(config);
+    config.name && item.setName(config.name);
+};
+
 export const furnish = scene => () => {
     const configs = scene.context.theme.furniture || [];
```
AdamXA, line 19: were you going to change this from furniture to items?

FostUK: just doing that in another PR. It's across all the configs so I thought I'd split it off.

PR #247, opened 2020-05-19
[CGPROD-2402] Remove particle emitters on disabled select screen buttons

AdamXA:
Disabled buttons already have their POINTER_UP listeners removed - hence there is a listener count is 0 check.

### DIFF
--- a/src/components/select/select-particles.js
+++ b/src/components/select/select-particles.js
@@ -7,7 +7,10 @@ import { gmi } from "../../core/gmi/gmi.js";
 
 const onPointerOut = (emitter, button) => button.on(Phaser.Input.Events.POINTER_OUT, () => emitter.stop());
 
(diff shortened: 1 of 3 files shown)


src/components/select/select-particles.js
```diff
@@ -7,7 +7,10 @@ import { gmi } from "../../core/gmi/gmi.js";
 
 const onPointerOut = (emitter, button) => button.on(Phaser.Input.Events.POINTER_OUT, () => emitter.stop());
 
+const isButtonEnabled = button => button.listeners(Phaser.Input.Events.POINTER_UP).length === 0;
+
 const createEmitter = (scene, button, config) => {
+    if (isButtonEnabled(button)) return;
```
FostUK, line 13: You can access button.config I think. 

Maybe...

but this works.

AdamXA: hmm it wasn't on there, is that part of the state? and is the state config set on the button config at all?

PR #253, opened 2020-06-03
CGPROD 2480 Single Item Select

FostUK:
* Disable next / previous buttons when only one page

### DIFF
--- a/src/components/select/select-screen.js
+++ b/src/components/select/select-screen.js
@@ -33,15 +33,18 @@ export class Select extends Screen {
     create() {
         this.addBackgroundItems();
         createTitles(this);
-        const buttons = ["home", "pause", "previous", "next"];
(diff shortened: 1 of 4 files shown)


src/components/select/select-screen.js
```diff
@@ -33,15 +33,18 @@ export class Select extends Screen {
     create() {
         this.addBackgroundItems();
         createTitles(this);
-        const buttons = ["home", "pause", "previous", "next"];
+        const onePage = this.config.choices.length > this.config.columns * this.config.rows;
+        const pagingButtons = onePage ? ["previous", "next"] : [];
```
AdamXA, line 37: if onePage is true, then it shows the previous and next buttons
think this should be renamed or the condition changed to check for one page instead of more than one page

FostUK: Oh yeah - that's totally backwards!

PR #254, opened 2020-06-03
CGPROD 2499 Audio skip

FostUK:
* Force end animations and audio when the skip button is pressed.

### DIFF
--- a/src/core/background/pages.js
+++ b/src/core/background/pages.js
@@ -24,7 +24,7 @@ const getTimedItems = scene => {
     return timedItems.map(createItems(scene));
 };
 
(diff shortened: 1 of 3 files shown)


src/core/background/pages.js
```diff
@@ -24,7 +24,7 @@ const getTimedItems = scene => {
     return timedItems.map(createItems(scene));
 };
 
-const skip = timedItems => timedItems.forEach(item => item.stop(1));
+export const skip = timedItems => timedItems.forEach(item => item.stop(1));
```
AdamXA, line 27: what does the 1 do?

FostUK: Tween stop(1) skips to the end of the tween.

PR #259, opened 2020-06-10
CGPROD-2459 Add source to pause level select btn

AnthonyGoodchild:
Adds param to stat for the level select button on pause screen which
contains level ID.

https://jira.dev.bbc.co.uk/browse/CGPROD-2459


AnthonyGoodchild:
mahalo

PR #261, opened 2020-06-10
Noticket debug measure

FostUK:
Capture

* Adds a measuring tool to debug mode


FostUK:
> is all this debug stuff getting included in every games code? can we remove it for production?
Currently the code is included but the assets are a separate load. 

Would be nice to be able to strip some of this out. Could maybe be done via a conditional import although that would only work in modern browsers...

PR #263, opened 2020-06-11
Cgprod 2458 level select

alex-s2:
Don't show level select button if already on level select

### DIFF
--- a/src/components/overlays/pause.js
+++ b/src/components/overlays/pause.js
@@ -17,9 +17,11 @@ export class Pause extends Screen {
     create() {
         this.addBackgroundItems();
         const parentKey = this.context.parentScreens.slice(-1)[0].scene.key;
+        const isAboveSelectScreen = parentKey.includes("select");
         const achievements = this.context.config.theme.game.achievements ? ["achievements"] : [];
         const pauseReplay = this.context.navigation[parentKey].routes.restart ? ["pauseReplay"] : [];
-        const levelSelect = this.context.navigation["pause"].routes.select ? ["levelSelect"] : [];
+        let levelSelect = this.context.navigation["pause"].routes.select ? ["levelSelect"] : [];
+        levelSelect = isAboveSelectScreen ? [] : levelSelect;
         const buttons = ["home", "audio", "settings", "pausePlay", "howToPlay"];
 
         this.setLayout([...buttons, ...achievements, ...levelSelect, ...pauseReplay]);
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 | Version | Description |
 |---------|-------------|
+| | Dont show level select button on pause screen when already on select screen. |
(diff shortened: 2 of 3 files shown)


src/components/overlays/pause.js
```diff
@@ -17,9 +17,11 @@ export class Pause extends Screen {
     create() {
         this.addBackgroundItems();
         const parentKey = this.context.parentScreens.slice(-1)[0].scene.key;
+        const isAboveSelectScreen = parentKey.includes("select");
```
AdamXA, line 20: I wonder if getting the scene (via https://photonstorm.github.io/phaser3-docs/Phaser.Scenes.SceneManager.html#getScene or some other route) and seeing if its an `instanceof Select` is a better way? Dunno :shrug: 😄 this seems fine too if the key always has select in.

AnthonyGoodchild, line 20: Do we only want it to show if you are not on _any_ select screen? i.e you could be on 3 of 5 select screens and you won't see the button to take you back to the first select screen?

AdamXA: saw this now, was thinking the same thing when I commented 😛

AnthonyGoodchild: I see the ticket suggests that the play button is for that purpose, I guess it may be confusing if you see a select screen button while you're on a select screen and it takes you to a different select screen (the first one)

alex-s2: Yeah it is possible to have that flow but this button itself only existed for like two weeks so it's not gonna be the end of the world plus the ticket says don't show it on a select screen so 🤷 😂

AnthonyGoodchild: yeah I think that's a requirement for select screens

alex-s2: Yeah it is a valid point but the ticket says dont show on select, plus this button itself has only existed for like two weeks and isnt in any game yet so its not like lost functionality

AnthonyGoodchild: `instanceof Select` might be clearer though

alex-s2: And yeah nick was saying it would be good to move to instanceof checking but a lot of things will have to be moved over to enable the christmas game so this seems ok for now.

AdamXA:
would just check with UX in post-amigo about how they think it should work... these buttons are weird 😛

AnthonyGoodchild:
@AdamXA 
> bit confused, is it not possible to have a game flow that looks like this:
> 
> home -> char select -> level select -> game
> 
> if I'm on level select, can I not have a select button that takes me back to character?
> I find it weird select screen use a home button and not a back button, then this wouldn't be an issue, could just click back instead... maybe that's a better way.

`select` destination is defined in routes, so you can send to 1st select screen (or any) in theory, as I understand.

Back button should definitely be a thing, maybe check with backlog as we've had this discussion.

PR #264, opened 2020-06-11
NOTICKET overlay stats screen

FostUK:
* Removes the code that stopped overlays calling set stats screen because it is now needed.


AnthonyGoodchild:
imsorry

PR #269, opened 2020-06-17
NOTICKET Repack tool

FostUK:
Repack Tool for converting Phaser 2 asset packs to 3.

### DIFF
--- a/dev/scripts/repack.js
+++ b/dev/scripts/repack.js
@@ -27,13 +27,13 @@ const audio = entry => {
     return entry;
 };
 
-const isDefined = value => value !== undefined
+const isDefined = value => value !== undefined;
 const spriteParams = ["frameWidth", "frameHeight", "margin", "spacing", "startFrame", "endFrame"];
 
 const spritesheet = entry => {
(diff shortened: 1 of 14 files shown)


dev/scripts/repack.js
```diff
@@ -0,0 +1,72 @@
...
+
+const isDefined = value => value !== undefined;
+const spriteParams = ["frameWidth", "frameHeight", "margin", "spacing", "startFrame", "endFrame"];
+
+const spritesheet = entry => {
+    const entries = spriteParams.map(param => [param, entry[param]]).filter(a => isDefined(a[1]));
+    entry.frameConfig = Object.fromEntries(entries);
+    [...spriteParams, "frameMax"].forEach(param => delete entry[param]);
+    return entry;
+};
```
AnthonyGoodchild, line 43: I feel like this could be a single array operation but I don't have enough coffee in my blood to follow up on that

FostUK: It's because "frameMax" isn't a thing in Phaser 3, do it needs deleting but doesn't need copying.

PR #270, opened 2020-06-17
 CGPROD-2516_Narrative-skip-button-duplicate-click 

Leonie-:
Prevents duplicate stat firing on Narrative screen "skip" button.

### DIFF
--- a/src/core/layout/gel-defaults.js
+++ b/src/core/layout/gel-defaults.js
@@ -204,7 +204,6 @@ export const config = screen => {
             id: "skip",
             channel: buttonsChannel(screen),
             action: ({ screen }) => {
-                gmi.sendStatsEvent("skip", "click");
                 screen.navigation.next();
             },
         },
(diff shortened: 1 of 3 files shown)


src/core/layout/gel-defaults.js
```diff
@@ -204,7 +204,6 @@ export const config = screen => {
             id: "skip",
             channel: buttonsChannel(screen),
             action: ({ screen }) => {
-                gmi.sendStatsEvent("skip", "click");
```
AnthonyGoodchild, line 207: will we not lose all skip stats with removing this?

Leonie-: Nope, it's not used anywhere else.

AnthonyGoodchild: yeah that's what I mean, I thought we were getting 2 skip stats being sent and that we do actually want a skip stat.

PR #284, opened 2020-07-06
CGPROD-2522 gelGrid param for showing select page

AnthonyGoodchild:
Adds gelGrid param for showing the page of a choice on load.

Used on select-screen via transientData

```
this.transientData["level-select"].choice = { id: "choice-id" };
```

https://jira.dev.bbc.co.uk/browse/CGPROD-2522

### DIFF
--- a/src/core/layout/grid/grid.js
+++ b/src/core/layout/grid/grid.js
@@ -33,6 +33,7 @@ export class GelGrid extends Phaser.GameObjects.Container {
     }
 
     addGridCells(theme) {
+        this.page = this.getCellPage(theme, this._config.showChoice);
         this._cells = theme.choices.map((cell, idx) => createCell(this, cell, idx, theme));
         this.makeAccessible();
         return this._cells;
@@ -155,6 +156,13 @@ export class GelGrid extends Phaser.GameObjects.Container {
         const pageMax = this.cellsPerPage * (pageNum + 1);
         const pageMin = this.cellsPerPage * pageNum;
         return this._cells.filter((cell, idx) => idx >= pageMin && idx < pageMax);
+    }
+
+    getCellPage(theme, cellId = 0) {
+        const cellIdx = theme.choices.reduce((cellIdx, cell, idx) => {
+            return cell.id === cellId ? idx : cellIdx;
+        }, 0);
+        return Math.floor(cellIdx / this.cellsPerPage);
     }
 
     reset() {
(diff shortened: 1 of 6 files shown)


src/core/layout/grid/grid.js
```diff
@@ -157,6 +158,13 @@ export class GelGrid extends Phaser.GameObjects.Container {
         return this._cells.filter((cell, idx) => idx >= pageMin && idx < pageMax);
     }
 
+    getCellPage(theme, cellId = 0) {
+        const cellIdx = theme.choices.reduce((cellIdx, cell, idx) => {
+            return cell.id === cellId ? idx : cellIdx;
+        }, 0);
```
FostUK, line 164: Would findIndex work for this?
`const cellIdx = Math.Max(theme.choices.findIndex(cell => cell.id === cellId), 0);`

?

src/core/layout/grid/grid.js
```diff
@@ -33,6 +33,7 @@ export class GelGrid extends Phaser.GameObjects.Container {
     }
 
     addGridCells(theme) {
+        this.page = this.getCellPage(theme, this._config.showChoice);
```
FostUK, line 36: Any reason why you aren't using the `transientData.choice` value here that was getting saved anyway?

AnthonyGoodchild: I've been over it again and while using `choice` does technically work, we would be saying that choice transient data does not necessarily reflect the choice which the player made.
 
If we are happy saying that then fine, I think it slightly muddies whatever the game does around the selected choice but any game changing this should be aware of what the value means anyway.

src/core/layout/grid/grid.js
AnthonyGoodchild: yeah not sure why I didn't do that

AnthonyGoodchild: I guess it was the default value, I should use `Math.max` in these kinds of logic more often, very complimentary.

src/core/layout/grid/grid.js
```diff
@@ -157,6 +158,14 @@ export class GelGrid extends Phaser.GameObjects.Container {
         return this._cells.filter((cell, idx) => idx >= pageMin && idx < pageMax);
     }
 
+    getCellPage(theme, cellId = 0) {
+        const cellIdx = Math.max(
+            theme.choices.findIndex(cell => cell.id === cellId),
+            0,
+        );
```
FostUK, line 165: Aww! Prettier! So annoying :)

PR #287, opened 2020-07-12
NOTICKET Gel scroll factor

FostUK:
* Sets scrollFactor to zero on gel root / groups / buttons so camera movement is ignored.
* Updates getbounds methods to return correct values.


AdamXA:
All looks good, but I guess the debug screen is offset 700, 300 still because it uses gel buttons and they are not in the offset root container?

FostUK:
Ah thanks. Will check. Should probably check this with Tower defence too as they use overlays.

PR #293, opened 2020-07-30
Replace @babel/polyfill with core-js

AdamXA:


AdamXA:
@babel/polyfill I removed, think it still wanted the other two, will double check.

AdamXA:
yea it still wants the others

PR #294, opened 2020-07-30
NOTICKET Corejs config

FostUK:
Prevent corejs from being transpiled by Babel


AdamXA:
Will shove this into 3.8.2 now

PR #298, opened 2020-09-02
[CGPROD-2682] string templating allows sprite selection depending on results

BeeGodwin:
- `ResultsSprite` takes a string template as a key composing an asset name with some value from transientData
- Debug screens can now be provided with a `prompt` object which pops up a window.prompt() taking a JSON string to be used for transientData
- Added a new debug screen, `results-conditional-sprites.json5`, where the player can be awarded 0, 1, 2, or 3 stars & the correct image displayed
- Updated tests


AnthonyGoodchild (approved):
looks good 👍😸👍

PR #299, opened 2020-09-03
[CGPROD-2708] results selects row audio asset based on template string

BeeGodwin:
- Similar to CGPROD-2682
- modified `results-row-audio.js` to accept a template string for audio assets and feed it a value from transientData
- modified & renamed `results-conditional-sprites.json5` to `results-conditional-assets.json5` and added a string templated asset key
- update tests to 100% coverage


src/core/debug/launcher.js
```diff
@@ -46,7 +46,7 @@ const getButtonConfig = launcher => (id, idx) => ({
 const getTransientData = example => {
     const transientData = example.transientData || {};
     if (!example.prompt) return transientData;
-    const response = JSON.parse(prompt(example.prompt.title, example.prompt.default));
+    const response = JSON.parse(prompt(example.prompt.title, example.prompt.default) || null);
```
AdamXA, line 49: how come it needs the extra null now?

AdamXA (approved):
looks great

src/core/debug/launcher.js
BeeGodwin: Think it actually needed it before tbh- it was crashing on hitting 'cancel' from the prompt.

PR #302, opened 2020-09-09
[CGPROD-2708] Bug fix- range clamp for conditional asset picking from transientData

BeeGodwin:
- We were crashing when passing a negative `keys` value and displaying a green box on a negative `stars` value
- Clamped the ranges passed from transientData so `stars` is in the range 0-3 and `keys` is in the range 0-1


FostUK (approved):
🗼

PR #303, opened 2020-09-09
[CGPROD-2708] fix silent audio on conditional assets debug screen

BeeGodwin:
- `window.prompt()` we added to parameterise the conditional assets screen steals focus in a browser-dependent way
- so, prevent the audio engine pausing on blur.


FostUK (approved):
Nice find!

Looks good. Might need a test to pass first.

PR #304, opened 2020-09-14
[CGPROD-2720] fix error spam on missing hover states for GEL buttons

BeeGodwin:
- GEL buttons now check that frame 1 exists before trying to set their sprite to frame 1
- Added unit test


AnthonyGoodchild (approved):
💯

PR #306, opened 2020-09-15
[CGPROD-2724] results debug fixes

BeeGodwin:
- Clearer prompt text for conditional asset screen
- Clamp gems range to min 0 for conditional asset screen
- Flip star assets so stars fill left to right


AnthonyGoodchild (approved):

AnthonyGoodchild:
Glass half full

PR #308, opened 2020-09-16
[CGPROD-2705] item registry

BeeGodwin:
- create item-registry.js which provides:
- itemsRegistry, which is a map of registries (for example, items, states, characters etc) that are needed beyond the Screen context
- `initRegistry` which takes a key and an array of items (example in themes/default/shop/config.json5) and returns a registry object, and also stashes it in the itemsRegistry map
- registry object has get(), set() and getCategory()
- get("item-id") returns item with a matching id or undefined
- getCategory("category") returns all items in category (items can be in more than one)
- set(id, diff) merges diff into the item with a matching id
- also add items.md documenting work so far.
- add tests.

This PR doesn't hook the registry into the game or anything- this will need to be done before we reach the shop screen but out of scope for this ticket.


themes/default/shop/config.json5
```diff
@@ -0,0 +1,24 @@
+{
```
FostUK, line 1: might want to add json5 to prettier config. Looks like it's doing 2 space indent here.

BeeGodwin: 76e3be6

src/components/shop/item-registry.js
```diff
@@ -0,0 +1,48 @@
...
+ * @author BBC Children's D+E
+ * @license Apache-2.0
+ */
+
+import fp from "../../../lib/lodash/fp/fp.js";
+
+export const initRegistry = itemsArray => {
+    const items = fp.clone(itemsArray);
+
+    const getResult = result => {
```
FostUK, line 12: Can use implicit return to make this one line if you prefer.

BeeGodwin: 5e992d4

FostUK (commented):
Looks good so far.
We are probably going to need some way to have more than one item list at some point,

Maybe we should try and have a quick end mob on this tomorrow?

Some minor things we can discuss or test in the future:
* We maybe don't need to worry too much about cloning things on init since they will be a one time load. It's good practice to be immutable but unlikely to have conflicts here. Having the bare minimum for now to get this up and running might be the safest course.
* I'm not sure what set is doing here but we don't need it for a while.
* Might be worth investigating maps since we want things to have unique ids.

src/components/shop/item-registry.js
```diff
@@ -0,0 +1,46 @@
...
+
+import fp from "../../../lib/lodash/fp/fp.js";
+
+export const initRegistry = itemsArray => {
+    const items = fp.clone(itemsArray);
+
+    const getResult = result => (result ? Object.assign({}, result) : undefined);
+
+    const get = (id, categories = []) => {
+        if (!id) return fp.clone(items);
```
AdamXA, line 15: what happens if id is 0 :D

src/components/shop/item-registry.js
```diff
@@ -0,0 +1,46 @@
...
+            result = items.find(item => id === item.id && item.category.includes(category));
+        });
+        return getResult(result);
+    };
+
+    const getCategory = category => items.filter(item => item.category.includes(category));
+
+    const set = item => {
+        const itemIndex = items.findIndex(storedItem => item.id === storedItem.id);
+        if (itemIndex >= 0) {
```
AdamXA, line 31: might be easier to read (or realise .findIndex returns -1 if nothing is found) if we were to check itemIndex is not -1 instead

src/components/shop/item-registry.js
```diff
@@ -0,0 +1,46 @@
+/**
+ * @copyright BBC 2020
+ * @author BBC Children's D+E
+ * @license Apache-2.0
+ */
+
+import fp from "../../../lib/lodash/fp/fp.js";
+
+export const initRegistry = itemsArray => {
+    const items = fp.clone(itemsArray);
```
AdamXA, line 10: it doesn't look like we're mutating itemsArray anywhere, so we _might_ be ok just referencing that rather than shallow copying it

src/components/shop/item-registry.js
AdamXA: though my page is only just showing nicks comments now and I realise now he may want us to use maps instead

src/components/shop/item-registry.js
BeeGodwin: ab34489

src/components/shop/item-registry.js
BeeGodwin: ab34489

src/components/shop/item-registry.js
BeeGodwin: ab34489

FostUK (approved):
Looks great! Nice job! 🐢

PR #310, opened 2020-09-18
[CGPROD-2706] item config auto loading

BeeGodwin:
- Renamed 'registry' to 'catalogue'
- Loader now loads catalogue
- configs can specify a catalogueKey: if they do, loader looks for a matching json5 in `themes/default/items`
- loader brings those json5s into the cache and uses them to initialise the catalogue as per CGPROD-2705
- updated docs, added tests.


src/core/loader/get-registry.js
```diff
@@ -0,0 +1,37 @@
...
+    });
+};
+
+const getRegistryKeys = config => {
+    const registryKey = item => item[1].registryKey;
+
+    return Object.entries(config).filter(registryKey).map(registryKey);
+};
+
+const loadToCache = (screen, key) => {
```
FostUK, line 32: I think, if you make this **const loadToCache = screen => key => {**

then you can do 

**registryKeys.forEach(loadToCache(screen);**

on line 13 above. Might be worth us changing that in the other bit of loader at some point...

src/core/screen.js
```diff
@@ -45,6 +45,7 @@ export class Screen extends Phaser.Scene {
             activeScreens,
             navigation: this._data.navigation,
             transientData: this._data.transient || {},
+            registry: this._data.registry,
```
FostUK, line 48: As discussed - this isn't required. In generally we also avoid sharing anything as much as possible via classes what with classes being terrible - unfortunately Phaser being class based makes this somewhat difficult to avoid.

src/core/loader/get-registry.js
```diff
@@ -0,0 +1,37 @@
+/**
+ * @copyright BBC 2020
+ * @author BBC Children's D+E
+ * @license Apache-2.0
+ */
+import { itemsRegistry, initRegistry } from "../../components/shop/item-registry.js";
+
+export const loadRegistry = (screen, config) => {
+    const registryKeys = getRegistryKeys(config);
+    if (!registryKeys.length) return;
```
FostUK, line 10: If there are no items in the config will it return an empty array? Prob a good pattern to do that.

You can delete this line if so.

BeeGodwin: It will, and you're right that the line isn't strictly needed, but early return avoids the need to modify the rest of the loader tests plus I thought it had semantic value. Happy to delete on the 'less is more' principle if you prefer.

src/core/screen.js
BeeGodwin: Yeah, meant to take this out. TY =)

src/core/loader/get-registry.js
BeeGodwin: I don't think I've seen (or, at least, understood) that double arrow pattern, but it's currying isn't it? Lovely little bit of functional goodness!

src/core/loader/get-registry.js
```diff
@@ -0,0 +1,37 @@
...
+            initRegistry(key, screen.cache.json.get(`registry-${key}`));
+        });
+        console.log("BEEBUG: itemsRegistry", itemsRegistry);
+    });
+};
+
+const getRegistryKeys = config => {
+    const registryKey = item => item[1].registryKey;
+
+    return Object.entries(config).filter(registryKey).map(registryKey);
```
FostUK, line 29: Is the filter trying to remove undefined entries?
Miht be clearer to map first then .filter(Boolean) ?

src/core/loader/get-registry.js
```diff
@@ -0,0 +1,37 @@
...
+    const registryKeys = getRegistryKeys(config);
+    if (!registryKeys.length) return;
+
+    registryKeys.forEach(key => {
+        loadToCache(screen, key);
+    });
+
+    screen.load.start();
+
+    screen.load.on("complete", () => {
```
AdamXA, line 18: define complete callback before starting loader

src/core/loader/get-registry.js
```diff
@@ -0,0 +1,37 @@
...
+ */
+import { itemsRegistry, initRegistry } from "../../components/shop/item-registry.js";
+
+export const loadRegistry = (screen, config) => {
+    const registryKeys = getRegistryKeys(config);
+    if (!registryKeys.length) return;
+
+    registryKeys.forEach(key => {
+        loadToCache(screen, key);
+    });
```
AdamXA, line 14: can be done in one line `registryKeys.forEach(key => loadToCache(screen, key));`

src/core/loader/get-registry.js
BeeGodwin: 672e4a9

src/core/screen.js
BeeGodwin: 672e4a9

src/core/loader/get-registry.js
BeeGodwin: Used the currying approach suggested by Nick in the comment above, per 672e4a9

src/core/loader/get-registry.js
BeeGodwin: I'm learning all _sorts_ of functional JS tricks today. TY!

src/core/loader/get-registry.js
BeeGodwin: On reflection, leaving it in is a code smell- it's better to make the needed changes in the loader tests (just adding a mock anyway.)

src/core/loader/get-catalogue.js
```diff
@@ -0,0 +1,33 @@
...
+ * @license Apache-2.0
+ */
+import { initCatalogue } from "../../components/shop/item-catalogue.js";
+
+export const loadCatalogue = (screen, config) => {
+    const catalogueKeys = getCatalogueKeys(config);
+
+    catalogueKeys.forEach(loadToCache(screen));
+
+    screen.load.on("complete", () => {
```
AdamXA, line 13: this only needs to be called once I guess, so you can use `.once` instead

AdamXA: could probably get rid of some of the `{ }` in this callback too

BeeGodwin: f84677e

docs/development/shop-component/items.md
```diff
@@ -1,27 +1,39 @@
...
-import { initRegistry } from "../core/item-registry.js";
+Your scene config should contain a `catalogueKey` which should be a string corresponding to the name of a .json5 file in `themes/default/items`. The loader will then pick them up and make them available as a map. 
+ 
 
-const registry = initRegistry("registryKey", ['array' , 'of', 'items'])); // tbd
+```javascript
+import { catalogue } from "./components/shop/item-catalogue.js";
+const catalogueSection = catalogue.get(sceneConfig.catalogueKey); // corresponding to a .json5 in items/
+const snotulonCannon = catalogueSection.get("Snotulon Cannon"); // get a single item
+const weapons = catalogueSection.getCategory(["weapons"s]); // filter by array of categories
```
AdamXA, line 25: typo

src/core/loader/get-catalogue.js
```diff
@@ -0,0 +1,31 @@
...
+        catalogueKeys.forEach(key => initCatalogue(key, screen.cache.json.get(`catalogue-${key}`))),
+    );
+
+    screen.load.start();
+};
+
+export const getCatalogueKeys = config => {
+    const catalogueKey = item => item[1].catalogueKey;
+
+    return Object.entries(config).map(catalogueKey).filter(Boolean);
```
AdamXA, line 23: can you just use `Object.values` instead of using entries to get the value? (finding it hard to follow what's happening here)

AdamXA: I'm guessing it could just end up as `Object.values(config).filter(Boolean);`

docs/development/shop-component/items.md
AdamXA: "weapon"s

src/core/loader/get-catalogue.js
BeeGodwin: 5a44dfc

src/components/shop/item-catalogue.js
```diff
@@ -26,14 +26,14 @@ export const initRegistry = (key, items) => {
         return false;
     };
 
-    const itemList = {
+    const catalogueSection = {
```
FostUK, line 29: Prefer itemList for this. Does what it says on the tin and this is the end agencies will deal with.

src/components/shop/item-catalogue.js
```diff
@@ -5,11 +5,11 @@
  */
 import fp from "../../../lib/lodash/fp/fp.js";
```
FostUK, line 6: Is this used still?

BeeGodwin: Yep, currently on line 23.

src/components/shop/item-catalogue.js
BeeGodwin: 9b5c1cb

AnthonyGoodchild (approved):
lgtm 🛒

PR #311, opened 2020-10-01
[CGPROD-2672] Scrollable list component

BeeGodwin:
- Based on the spike code for CGPROD-2722
- Brings in the minified rexUI bundle
- Adds a stub for the Shop and adds it to main.js (you'll need to hook it up to something's next() to see it)
- Creates a scrollable list with dummy data
- Establishes some of the config.json5 needed for the shop- i.e. the JSON blobs for overlays (the component parts of the item buttons.)
- Some assumptions about item config are made here (items here have an id, name, description, icon, price, and ariaLabel.)

There's a known issue here where the list doesn't render exactly the same on mobile & non-mobile breakpoints- this might need some changes making to gel-button.js, but think those should be a separate concern. 

Planning / expecting to refactor some of this, but raising the PR early as it's a bit large.


src/main.js
```diff
@@ -9,13 +9,16 @@ import { Results } from "./components/results/results-screen.js";
...
+import { Shop } from "./components/shop/shop.js";
 import { Pause } from "./components/overlays/pause.js";
 import { settingsChannel } from "./core/settings.js";
 import { eventBus } from "./core/event-bus.js";
 import { startup } from "./core/startup.js";
+
 //TODO Re-enable once Phaser plugins are built (should be in 3.24)
 //import "/node_modules/phaser/plugins/spine/dist/SpinePlugin.min.js";
 import "../lib/SpinePlugin.min.js"; //CAN BE REMOVED IF NOT USING SPINE
+import "../lib/rexuiplugin.min.js"; // CAN BE REMOVED IF NOT USING SHOP
```
AdamXA, line 21: could this be imported in the shop so it only gets used when shop component is imported?

BeeGodwin: I'll try that but I suspect it needs to be imported before game instantiation as it's a plugin (see get-default-plugins.js.)

src/core/layout/gel-button.js
```diff
@@ -19,10 +19,11 @@ export class GelButton extends Phaser.GameObjects.Container {
     constructor(scene, x, y, config) {
         const metrics = getMetrics();
         super(scene, x, y);
-
         this.sprite = scene.add.sprite(0, 0, assetPath(Object.assign({}, config, { isMobile: metrics.isMobile })));
         this.setScrollFactor(0);
-        this.sprite.setScrollFactor(0);
+        if (!config.inScrollable) {
```
FostUK, line 24: Might be nice to make this more generic sounding rather than tying its name to the scrollable list - I think we might need it elsewhere so maybe just "scrollable" ?

Also feel free to use shortcutting to keep the line count down like:

        `!config.scrollable && this.sprite.setScrollFactor(0);`

src/core/layout/scrollable-list/scrollable-list.js
```diff
@@ -0,0 +1,94 @@
+/**
```
FostUK, line 1: Just an observation - but does everything in here need to be exported?
I recently edited the catalogue as there was only one thing needed exporting.

We tend to only export what is needed - exports being the public interface and eveything else being private.

If all this is neede though that's fine :)

src/main.js
FostUK: We should probably talk about the ES6 module thing - will try to find a window tomorrow. If it was an ES6 module it would be easier I think as we support zero build loading- I think right now you are possibly using our global loader to do this (I haven't looked though ) which would mean it has to be loaded up front. There's Phaser examples of loading a plugin in the scene's preloader so in theory this should be possible.

src/core/layout/scrollable-list/scrollable-list.js
```diff
@@ -0,0 +1,94 @@
...
+    return {
+        x: 0,
+        y: 0,
+        width: safeArea.width,
+        height: safeArea.height,
+        scrollMode: 0,
+        background: scene.add.image(0, 0, assetKey(keys.background, keys)),
+        panel: {
+            child: lib.createPanel(scene),
+            mask: {
```
FostUK, line 31: Totally minor but some of this can maybe be squashed up. Prettier will leave it alone if it fits on one line...

src/core/layout/scrollable-list/scrollable-list.js
```diff
@@ -0,0 +1,94 @@
...
+    return scene.rexUI.add
+        .sizer({
+            orientation: "y",
+            space: { left: 10, right: 10, top: 0, bottom: 10, item: 10 },
+        })
+        .add(table, 1, "center", 0, true);
+};
+
+export const createItem = (scene, item) => {
+    const label = scene.rexUI.add.label({
```
FostUK, line 79: You can implicit return this and save a line or two :)

src/main.js
BeeGodwin: Yes, let's talk further. For now, 4e4fd72 demonstrates we can happily move the import to the Shop screen and de-bloat the bundle somewhat for non-Shop games (as Adam suggested.)

src/core/layout/scrollable-list/scrollable-list.js
BeeGodwin: Fixed in today's refactor- now it's only exporting the interface.

src/components/shop/shop.js
```diff
@@ -0,0 +1,28 @@
...
+ * @copyright BBC 2020
+ * @author BBC Children's D+E
+ * @license Apache-2.0
+ */
+import { Screen } from "../../core/screen.js";
+import { scrollableList } from "../../core/layout/scrollable-list/scrollable-list.js";
+import RexUIPlugin from "../../../lib/rexuiplugin.min.js";
+
+export class Shop extends Screen {
+    preload() {
```
FostUK, line 14: I'm a bit unsure about this format over using `this.plugins.installScenePlugin('rexUI', RexUIPlugin, "rexUI", this )`. It (supposedly) adds a second load call - which we should try and avoid/is uncesscary because it's been compiled in via an import). I honestly am unsure how this even can be working in a final build.

`installScenePlugin` (supposedly) just installs the plugin code.
Could see if it's loading it from `lib` on live.

themes/default/shop/assets.json
```diff
@@ -0,0 +1,63 @@
+{
+    "results": {
```
AdamXA, line 2: results

BeeGodwin: 🤦‍♀️ good spot

src/core/layout/scrollable-list/button-overlays.js
```diff
@@ -0,0 +1,36 @@
+/**
+ * @module core/layout/scrollable-list
+ * @copyright BBC 2020
+ * @author BBC Children's D+E
+ * @license Apache-2.0 Apache-2.0
+ */
+import fp from "../../../../lib/lodash/fp/fp.js";
+
+export const overlays1Wide = ({ scene, gelButton, item, config }) => {
```
FostUK, line 9: Super minor but we usually put exports at the bottom.
Don't think this is written down anywhere.  Might install an eslint plugin for this.

src/components/shop/shop.js
BeeGodwin: Both approaches work (I was failing on not passing `this` before), so happy to go the `installScenePlugin` route.

src/core/layout/scrollable-list/button-overlays.js
BeeGodwin: 33026bf

src/components/shop/shop.js
BeeGodwin: 33026bf

src/core/layout/scrollable-list/scrollable-list-buttons.js
```diff
@@ -0,0 +1,45 @@
...
+ * @author BBC Children's D+E
+ * @license Apache-2.0 Apache-2.0
+ */
+import { onClick } from "./scrollable-list-helpers.js";
+import { eventBus } from "../../event-bus.js";
+import { overlays1Wide } from "./button-overlays.js";
+import fp from "../../../../lib/lodash/fp/fp.js";
+
+export const createGelButton = (scene, item) => {
+    const id = `shop_id_${item.id}`;
```
FostUK, line 13: Might make a note somewhere that we should probably look at this. We can't always rely on a hard-coded word like "shop" because the scene name can be changed. I don't think we need to worry about this right now though - just mentioning it incase it comes up in later shop work

src/core/layout/scrollable-list/scrollable-list-buttons.js
```diff
@@ -0,0 +1,45 @@
...
+    eventBus.subscribe({
+        callback: () => onClick(gelButton),
+        channel: gelConfig.channel,
+        name: id,
+    });
+
+    return fp.flow(scaleButton, overlays1Wide)({ scene, gelButton, config, item });
+};
+
+export const scaleButton = args => {
```
FostUK, line 39: Does this need to be exported?

src/core/layout/scrollable-list/scrollable-list-buttons.js
BeeGodwin: I'll make it more generic for now- perhaps later we can set it from some config value.

src/core/layout/scrollable-list/scrollable-list-buttons.js
BeeGodwin: Ah! It does not. <toodles off to rewrite test>

FostUK (approved):
Great work btw - quite a long going on in this PR!

Approved for when you are ready.

src/core/layout/scrollable-list/scrollable-list-buttons.js
BeeGodwin: ab27f58

src/core/layout/scrollable-list/scrollable-list-buttons.js
BeeGodwin: ab27f58

PR #312, opened 2020-10-02
[CGPROD-2802] Fix crash on pressing enter key on how to play screen

AdamXA:
The how to play screen has that "accessible carousel" thing that has a weird screen reader only div on it that reads out what page you are on. It doesn't have a click event on it so it crashes.


AnthonyGoodchild (approved):
`pr.approve && pr.approve(this.diff.toReviewString())`

PR #314, opened 2020-10-08
[CGPROD-2795] responsive scrollable list

BeeGodwin:
- Scrollable list now consistently scales to fit the safe area on resize (if it's not right, then we need to update safe area)
- Inner padding is now consistent across any & all sizes
- Resolved a race condition that sometimes broke layout on resize
- Buttons now check that their click was within panel bounds before handling (gel buttons didn't respect rexUI's masking.)

### DIFF
--- a/src/core/layout/scrollable-list/scrollable-list-buttons.js
+++ b/src/core/layout/scrollable-list/scrollable-list-buttons.js
@@ -4,7 +4,7 @@
  * @author BBC Children's D+E
  * @license Apache-2.0 Apache-2.0
  */
-import { onClick } from "./scrollable-list-helpers.js";
+import { handleIfVisible } from "./scrollable-list-helpers.js";
 import { eventBus } from "../../event-bus.js";
 import { overlays1Wide } from "./button-overlays.js";
 import fp from "../../../../lib/lodash/fp/fp.js";
@@ -26,8 +26,9 @@ const createGelButton = (scene, item) => {
     };
 
     const gelButton = scene.add.gelButton(0, 0, gelConfig);
+    // console.log('BEEBUG: gelButton', gelButton);
     eventBus.subscribe({
-        callback: () => onClick(gelButton, scene),
+        callback: () => handleIfVisible(gelButton, scene),
         channel: gelConfig.channel,
         name: id,
     });
--- a/src/core/layout/scrollable-list/scrollable-list-buttons.js
+++ b/src/core/layout/scrollable-list/scrollable-list-buttons.js
@@ -7,7 +7,6 @@
 import { handleIfVisible } from "./scrollable-list-helpers.js";
 import { eventBus } from "../../event-bus.js";
 import { overlays1Wide } from "./button-overlays.js";
-import { accessibilify } from "../../../core/accessibility/accessibilify.js";
 import fp from "../../../../lib/lodash/fp/fp.js";
 
 const createGelButton = (scene, item) => {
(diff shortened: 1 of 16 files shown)


test/core/layout/scrollable-list/scrollable-list.test.js
```diff
@@ -40,75 +47,94 @@ const mockScene = {
     layout: {
         getSafeArea: jest.fn().mockReturnValue({ width: 100, height: 100 }),
     },
+    scale: { on: jest.fn() },
 };
 const mockGelButton = { width: 100, setScale: jest.fn() };
 buttons.createGelButton = jest.fn().mockReturnValue(mockGelButton);
+buttons.scaleButton = jest.fn();
 
 describe("Scrollable List", () => {
```
BeeGodwin, line 56: Urgh, changing the nesting has brought most of this test suite into the diff.

src/core/layout/scrollable-list/scrollable-list-buttons.js
```diff
@@ -39,9 +39,9 @@ const createGelButton = (scene, item) => {
...
     const { scene, config, gelButton } = args;
     const safeArea = scene.layout.getSafeArea();
-    const scaleFactor = (safeArea.width - config.space * 2) / gelButton.width;
+    const scaleFactor = (safeArea.width - config.space * 4) / gelButton.width;
     gelButton.setScale(scaleFactor);
     return args;
 };
 
-export { createGelButton };
+export { createGelButton, scaleButton };
```
AdamXA, line 47: @FostUK are we doing exports like this or `export const`?

AnthonyGoodchild: it's the same thing, no? just different syntax

FostUK: That's only one export that's an object with createGelButton and scaleButton methods.

FostUK: Kinda weird that it was an object with one method before :)

src/core/layout/scrollable-list/scrollable-list-helpers.js
```diff
@@ -4,12 +4,21 @@
  * @author BBC Children's D+E
  * @license Apache-2.0 Apache-2.0
  */
-
 /* eslint-disable no-console */
-export const onClick = gelButton => {
-    console.log(`Clicked ${gelButton.config.id}`); // placeholder
-};
 
-export const assetKey = (key, assetKeys) => {
```
FostUK, line 13: aNY REASON YOU ARE EXPORTING THIS TWICE?

src/core/layout/scrollable-list/scrollable-list-helpers.js
```diff
@@ -4,12 +4,21 @@
...
-    return [assetKeys.prefix, key].join(".");
+/* gelButton's clickable zone extends beyond rexUI's mask [CGPROD-2795]*/
```
FostUK, line 9: Prob don't need this comment - if there's anything that's a bit of a workaround we usually make sure there's a described test for it.

FostUK (approved):
Looks great! Added a minor comment.
Also **For future ref:** we try to avoid "helpers" files as they are a tiny bit of a code smell due to not really being descriptive and attract "code lint. You may have seen this somewhere else but it was a previous grad's work :)

Don't worry about it here though - merge merge merge!

PR #316, opened 2020-10-13
[CGPROD-2797] safe area debug

BeeGodwin:
- Provides a new key ('U') in debug mode that shows the safe area
- Adds `safe-area-draw.js` to get and draw this safe area
- Adds a new param `mirrorY` to `scene.layout.getSafeArea()`, defaulting to true, which has the safe area vertically centred (the way we want it in Select) 
- Uses mirrorY = false when getting the safe area for debug mode

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 | Version | Description |
 |---------|-------------|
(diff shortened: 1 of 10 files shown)


FostUK (approved):
Awesome!

🐑🐏🐄🐖🦙🚜

PR #317, opened 2020-10-13
[CGPROD-2763] draft

BeeGodwin:
- adds handler functions to the scrollable list
- runs the list items through accessibilify and provides a minimal custom group to hook the resulting a11y elems into existing resize mechanisms
- modify accessibilify to add the resize function that gets called on GEL buttons to the buttons themselves
- scrollable list conforms to updated safe area
- click masking conforms to updated safe area
- on scrolling, handler updates size and position of all GEL button a11y elems
- on tabbing through items, handler updates scroller if item tabbed to is off the top / off the bottom of the visible area


src/core/layout/scrollable-list/scrollable-list-buttons.js
```diff
@@ -32,16 +33,20 @@ const createGelButton = (scene, item) => {
...
-    const safeArea = scene.layout.getSafeArea();
+    const safeArea = scene.layout.getSafeArea({}, false);
     const scaleFactor = (safeArea.width - config.space * 4) / gelButton.width;
     gelButton.setScale(scaleFactor);
     return args;
 };
 
+const makeAccessible = args => {
+    accessibilify(args.gelButton);
+    return args;
```
AnthonyGoodchild, line 49: nitpick but I don't think this needs to return anything, does it?

BeeGodwin: correct, & good spot- it did at one point but now the only reason to return args is to keep it in the `fp.flow` call above. Will refactor.

src/core/accessibility/accessibilify.js
```diff
@@ -65,7 +67,9 @@ export function accessibilify(button, gameButton = true) {
         if (!button.input.enabled) return;
         button.emit(Phaser.Input.Events.POINTER_UP, button, sys.input.activePointer, false);
     };
-    const onMouseOver = () => button.emit(Phaser.Input.Events.POINTER_OVER, button, sys.input.activePointer, false);
+    const onMouseOver = () => {
+        button.emit(Phaser.Input.Events.POINTER_OVER, button, sys.input.activePointer, false);
```
FostUK, line 71: does this work as one line?

src/core/layout/scrollable-list/scrollable-list.js
```diff
@@ -4,48 +4,54 @@
  * @author BBC Children's D+E
  * @license Apache-2.0 Apache-2.0
  */
-import { assetKey } from "./scrollable-list-helpers.js";
+import { updatePanelOnFocus, updatePanelOnScroll } from "./scrollable-list-handlers.js";
 import { createGelButton, scaleButton } from "./scrollable-list-buttons.js";
+import * as a11y from "../../accessibility/accessibility-layer.js";
 import fp from "../../../../lib/lodash/fp/fp.js";
 
+const GRID_NAME = "grid";
```
FostUK, line 12: Could probably just remove this and rely on "grid" below. 
I get why this is a common pattern - you remove any problems with mis-spelling but it immediately stops all the code being portable.

src/core/layout/scrollable-list/scrollable-list-handlers.js
```diff
@@ -0,0 +1,75 @@
...
+    const lower = idx * rexLabel.height + idx * panel.space.top;
+    return { lower, upper: lower + rexLabel.height };
+};
+
+const updateScrollPosition = (panel, offset) => {
+    const maxOffset = getMaxOffset(panel);
+    const tDelta = offset / maxOffset;
+    const newT = panel.t + tDelta;
+    const tThreshold = 1 / getPanelItems(panel).length;
+    fp.cond([
```
FostUK, line 63: This probably needs naming something that makes sense - currently hard to tell what this does. The inevitable difficulty with fp :)

If you're immediately invoking cond then you're (usually) in a place where you need to name the function.

FostUK: You could also move this outsiode the current body -
```
const setThreshHoldFn = (tThreshold, panel) => fp.cond([
        [t => t < tThreshold, () => panel.setT(0)],
        [t => t > 1 - tThreshold, () => panel.setT(1)],
        [() => true, t => panel.setT(t)],
    ])(newT);
```

then do
```
const setTheshold =  setThreshHoldFn(tThreshold, panel);
setTheshold (newT)
```

Assuming this sets the threshold? Unsure...

src/core/layout/scrollable-list/scrollable-list-handlers.js
```diff
@@ -0,0 +1,75 @@
...
+const getPanelItems = panel => panel.getByName(GRID_NAME, true).getElement("items");
+
+const updatePanelOnFocus = panel => rexLabel => {
+    const visibleBounds = getVisibleRangeBounds(panel);
+    const itemBounds = getItemBounds(panel, rexLabel);
+    fp.cond([
+        [(vb, ib) => ib.lower < vb.lower, (vb, ib) => updateScrollPosition(panel, ib.lower - vb.lower)],
+        [(vb, ib) => ib.upper > vb.upper, (vb, ib) => updateScrollPosition(panel, ib.upper - vb.upper)],
+        [() => true, () => {}],
+    ])(visibleBounds, itemBounds);
```
AnthonyGoodchild, line 37: This feels like it could be written a bit more cleanly as a function. I'm not a huge fan of lodash but the 2 conds looks very similar - so could be possible to do the same operation on a conditional, given the 2 params.

AnthonyGoodchild: (feel free to ignore my bias, commenting for potential discussion)

src/core/layout/scrollable-list/scrollable-list-handlers.js
```diff
@@ -0,0 +1,75 @@
...
+const getItemBounds = (panel, rexLabel) => {
+    const idx = getPanelItems(panel).findIndex(item => item === rexLabel);
+    const lower = idx * rexLabel.height + idx * panel.space.top;
+    return { lower, upper: lower + rexLabel.height };
+};
+
+const updateScrollPosition = (panel, offset) => {
+    const maxOffset = getMaxOffset(panel);
+    const tDelta = offset / maxOffset;
+    const newT = panel.t + tDelta;
```
AdamXA, line 61: im a bit confused what the t and tThreshold are

src/core/layout/scrollable-list/scrollable-list-handlers.js
```diff
@@ -0,0 +1,75 @@
...
+    const tThreshold = 1 / getPanelItems(panel).length;
```
AdamXA, line 62: im a bit confused what the t and tThreshold are

src/core/layout/scrollable-list/scrollable-list-handlers.js
```diff
@@ -0,0 +1,75 @@
...
+ * @author BBC Children's D+E
+ * @license Apache-2.0 Apache-2.0
+ */
+/* eslint-disable no-console */
+
+import fp from "../../../../lib/lodash/fp/fp.js";
+
+const GRID_NAME = "grid";
+
+const handleIfVisible = (gelButton, scene) => {
```
FostUK, line 13: Bit unsure what this does exactly. Is it only called on a settings change? It might also be worth it returning a function rather than wrapping it in an anonymous fn later.

FostUK (approved):
Put a few comments. If you aren't in a hurry to commit - might be worth going over them together but will have to be tomorrow.

If you need to commit for further work though totally go ahead - looks great and anything small can always be refactored later.

src/core/accessibility/accessibilify.js
BeeGodwin: It does. Will fix.

src/core/layout/scrollable-list/scrollable-list.js
BeeGodwin: yeah- there's only the smallest semantic gain by making this a `const`. Code style thing I guess, last team preferred to `const` all strings. Will change.

src/core/layout/scrollable-list/scrollable-list-handlers.js
BeeGodwin: It checks on click that the button is in visible bounds, i.e. did we click on a button outside of the list's viewable area. This part of the code is incomplete as it just calls a placeholder click handler, but you're right, returning a function would be better. Will refactor.

src/core/layout/scrollable-list/scrollable-list-handlers.js
BeeGodwin: That's a really helpful example- thanks! Will refactor.

src/core/layout/scrollable-list/scrollable-list.js
BeeGodwin: 44416fa

src/core/accessibility/accessibilify.js
BeeGodwin: 44416fa

src/core/layout/scrollable-list/scrollable-list-handlers.js
```diff
@@ -0,0 +1,75 @@
...
+    const panel = gelButton.rexContainer.parent.getTopmostSizer();
+    const safeArea = scene.layout.getSafeArea({}, false);
+    const height = scene.scale.displaySize.height;
+    const topY = height / 2 + safeArea.y + panel.space.top;
+    const bottomY = topY + panel.innerHeight;
+    const mouseY = scene.input.y;
+    if (mouseY >= topY && mouseY <= bottomY) onClickPlaceholder(gelButton);
+};
+
+const onClickPlaceholder = gelButton => console.log(`Clicked ${gelButton.config.id}`);
```
AnthonyGoodchild, line 21: is this in some requirements? (console logging)

src/core/layout/scrollable-list/scrollable-list-handlers.js
BeeGodwin: a163e18

src/core/layout/scrollable-list/scrollable-list-handlers.js
BeeGodwin: I've just refactored this out- needed a placeholder click handler while working on it.

BeeGodwin: a163e18

src/core/layout/scrollable-list/scrollable-list-handlers.js
BeeGodwin: Naming improved. 08cc562

src/core/layout/scrollable-list/scrollable-list-handlers.js
BeeGodwin: t is the scrollbar position expressed as a value between 0 and 1 and the threshold thing is there to clamp t to 0 or 1 if we're scrolling to the first or last item. It was a bit confusing so I've renamed things... 08cc562

src/core/layout/scrollable-list/scrollable-list-handlers.js
BeeGodwin: 08cc562

PR #318, opened 2020-10-13
CGPROD 2707 remove states

FostUK:
Cool Robot

* Remove states system in favour of collections.
* Allow collections to supply an inline catalogue

To discuss:
* Rename "items" folder to collections? 
* Apply any defaults - not just `qty`


Leonie-:
+10 points for good GIF.

src/core/collections.js
```diff
@@ -12,18 +12,21 @@ export let collections = new Map();
 
 const mergeDefaults = itemDefaults => item => ({ ...item, ...itemDefaults?.find(def => def.id === item.id) });
 const getFilterParams = config => ({ ids: config.defaults?.map(x => x.id) ?? [], tags: config.include ?? [] });
-const filterItems = config => item => config.ids.includes(item.id) || item.tags.some(tag => config.tags.includes(tag));
-const addQty = qty => item => ({ ...item, qty });
+const tagIn = config => tag => config.tags.includes(tag);
+const include = config => item => !config.tags.length || config.ids.includes(item.id) || item.tags?.some(tagIn(config));
+const addQty = qty => item => ({ ...item, qty }); //TODO this should be add global defaults. qty might not even be present? needed for shop
```
BeeGodwin, line 17: It feels a little bit odd for states to have a quantity. I know it's not really going to hurt us (after all, there's still one of a unique thing even if you wouldn't usually think of that thing as existing as a quantity) so it might just be a question of tolerating a bit of semantic dissonance in the name of simplicity....

src/core/layout/layout.js
```diff
@@ -53,6 +53,7 @@ export function create(scene, metrics, buttonIds, accessibleButtonIds) {
     root.setScrollFactor(0);
 
     const addCustomGroup = (key, group, pos = 0) => {
+        debugger;
```
BeeGodwin, line 56: debugger

BeeGodwin (approved):
Ran locally and confirmed states still work as intended & tests all pass. Two small comments, but nothing blocking from me.

src/core/layout/layout.js
FostUK: Ha!

src/core/collections.js
FostUK: Yeah totally agree. I had 2 things to discuss on this:
* Rename "items" folder to collections?
* Apply any defaults - not just qty

I think the defaults should just be anything and then you could miss them out for select screens and so on.
Gonna stick that in another small follow up ticket though as this was mainly about removing the states file.
Prob needs some docs updates too come to think of it...

PR #321, opened 2020-10-19
[CGPROD-2763] bugfix- fix scaleButton call

BeeGodwin:
Managed to introduce a bug while refactoring. _cries in TDD_


AdamXA (approved):
dunno what a label is and what its children are, but looks good! :joy:

AnthonyGoodchild (approved):
Better working than not 👏

PR #322, opened 2020-10-20
[CGPROD-2764] shop boilerplate

BeeGodwin:
- Add home and pause buttons to boilerplate screen with appropriate flow through accessible elements on tabbing
- Updates shop config to include params for title & wallet UI - assets, strings, & font styling
- Creates two containers for shop & wallet UI elements and populates them
- Sizes and positions these elements based on the available space at the top of the screen (uses safe area & various metrics)
- Handles behaviour of these elements on resize
- Wallet resizes to a consistent overall size, elems inside the container shrink-to-fit
- Title resizes for a consistent font size, background image expands to fit
- UX tweak to allow y-axis padding to be different to x-axis padding (out of scope for the JIRA ticket but it gets in cos it's small)


FostUK (commented):
Looks good.
* We prob need to move as much out of the class as possible. getX getY are obvious candidates.
* Title - we have similar code in two places now. We should probably look into this and see what we need to do about it. Also worth double checking we can't just use the bacground items system.
* Wallet - probably should be in its own file. Also will then make sense not to preface everything with "wallet" :)

src/components/shop/shop-layout.js
```diff
@@ -0,0 +1,30 @@
...
+
+export const getYPos = (metrics, safeArea) => {
+    const { verticals, verticalBorderPad } = metrics;
+    const padding = (safeArea.y - verticals.top) / 2 + verticalBorderPad / 2;
+    return verticals.top + padding;
+};
+
+export const getScaleFactor = args => {
+    const { metrics, container, fixedWidth, safeArea } = args;
+    container.setScale(1);
```
AdamXA, line 22: dunno if a get function that sets the scale to 1 might cause confusing behaviour - I know this is used to set it afterwards anyway. is there maybe another way to get the width and height of the container? (can't remember been a while since I last looked at Phaser :D)

src/components/shop/shop.js
```diff
@@ -17,8 +21,43 @@ export class Shop extends Screen {
 
     create() {
         this.addBackgroundItems();
-        const buttons = ["exit", "audio"];
-        this.setLayout(buttons);
-        this.panel = scrollableList(this);
+        this.setLayout(["home", "pause"]);
+        const metrics = getMetrics();
+        this.title = this.createTitle(metrics);
+        this.wallet = createWallet(this, metrics);
```
FostUK, line 27: Is there a better name than wallet?
Wallet implies it holds the currency and deals with transactions where as this is just a ui display of currency.

I admit I am struggling to think of anything specific enough myself. Might be worth floating in game-dev.

src/core/layout/scrollable-list/scrollable-list.js
```diff
@@ -110,13 +96,24 @@ const setupEvents = (scene, panel) => {
...
-};
+        this.add(this.panel);
+        scene.layout.addCustomGroup(scene.scene.key, this, 0);
+        a11y.addGroupAt(scene.scene.key, 0);
+
+        scene.input.topOnly = false;
+        setupEvents(scene, this.panel);
+    }
+
+    reset() {
```
FostUK, line 112: You can replace this with 
`
const resizePanel = (scene, panel) => () => {
...

in constructor:
this.reset = resizePanel(this, panel);
```

this should allow you to remove `this.panel` from the class props.

File: src/components/shop/shop-layout.js (hunk 1 above)
Comment (BeeGodwin) [reply src/components/shop/shop-layout.js:22]: Maybe I need to use the scale to normalise the rest of the values. It does feel a bit dirty. Will look.

File: src/components/shop/shop.js (hunk 2 above)
Comment (BeeGodwin) [reply src/components/shop/shop.js]: Following chat, renamed `wallet` to `balance` across the PR. d856e23

File: src/main.js [hunk 4]
```diff
@@ -57,7 +57,7 @@ const screens = {
             debug: "debug",
             //Example of custom routing function
             next: scene => {
-                scene.navigate("narrative");
+                scene.navigate("shop");
```
FostUK, line 60: We probably don't want to leave this in for now. Might be time to set up a debug screen. Maybe as part of whatever is next?

FostUK (approved):
Looks great! Couple of small comments...

src/core/layout/scrollable-list/scrollable-list.js
BeeGodwin: I thumbs-up'd this without really understanding what you meant and have since had a think about it- I'll try it and you can tell me if I've made sense of it 😹

src/main.js
BeeGodwin: Yeah, this needs to come out.

src/components/shop/shop-layout.js
BeeGodwin: this was a good shout- I was able to dispense with the `.setScale` call. b283938

src/main.js
BeeGodwin: b283938

PR #323, opened 2020-10-22
CGPROD 2850 4/3 Area buttons

FostUK:
* Buttons within the 4/3 area now sit exactly on the edge in mobile breakpoint.
* Buttons will move inwards if the screen width causes the border pad to overlap the 4/3 area.

Mobile mode wide:
image

Mobile mode 4/3:

image

Old mobile mode:
image


alex-s2 (approved):
Looks good

PR #326, opened 2020-10-27
[CGPROD-2858] add shop debug screen

BeeGodwin:
- Add an example shop screen, accessible when `debug=true`


FostUK (approved):
looks great! 🐻‍❄️

PR #327, opened 2020-10-27
[CGPROD-2674] fix resize being a couple of px off

BeeGodwin:
- in post-flight on CGPROD-2674 it was noted that the scrollable list was not conforming to the edges of the safe area
- in a previous UX chat we noted X- and Y-axis padding values should be different to conform to the UX spec
- following the change we were not conforming because we were using a Y where we should have used an X.
- fixed.


AdamXA (approved):
incredible code change

PR #329, opened 2020-10-27
[CGPROD-2858] shop debug screen - unbreak asset paths, the fast way

BeeGodwin:
- Previous PR on this ticket broke all the asset paths as they relied on the shop existing in the theme and main.js
- Ideally we'd have all debug assets under the debug assets folder. however:
- gel buttons depend on a dot-separated asset path
- debug mode depends on underscore-separated paths
- hilarity ensues
- as a quick fix, add Shop back to main.js

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -107,7 +107,7 @@ const screens = {
         scene: Shop,
         routes: {
             home: "home",
-        }
+        },
     },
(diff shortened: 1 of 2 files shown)


AnthonyGoodchild (approved):
🍏

PR #331, opened 2020-10-28
[CGPROD-2848] fix debug measure tool

BeeGodwin:
- The scrollable list component relies on scene.input.topOnly === false
- Setting this broke the debug measuring tool for screens featuring the scrollable list
- This code just changes scene.input.topOnly to true when the measuring tool is visible
- It has the slightly undesirable side effect of defeating the scroll and click behaviours of the list while the measure tool is active.
- As it stands, it seems like the two features are mutually exclusive- but probably acceptable in this limited case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 | Version | Description |
 |---------|-------------|
 | | Add shop demo screen in debug mode |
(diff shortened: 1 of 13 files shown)


FostUK (approved):
Looks great! Good solution too. 🌠

PR #332, opened 2020-10-29
CGPROD 2771 Shop Transactions

FostUK:


collections-plan.md
```diff
@@ -0,0 +1,102 @@
...
+* Player buys shield
+* Shield is added to inventory
+* Shield shows in shop as “owned”
+
+**start collection states:**
+shop [{key: "shield"}]
+inventory []
+
+**Command**
+`collections.moveItem({key: "shield", from: "shop", to: "player", fromState: "owned"])`
```
AdamXA, line 38: I guess this is meant to say `to: "inventory"`

I understand that fromState changes the state of the thing the item came from, just found it confusing at first

collections-plan.md
```diff
@@ -0,0 +1,102 @@
...
+* Player buys 1 banana
+* 1 banana is added to inventory
+* 4 bananas show in shop
+
+**start collection states:**
+shop [{key: "banana", qty: 5}]
+inventory []
+
+**Command**
+`collections.moveItem({key: "banana", from: "shop", to: "player"})`
```
AdamXA, line 55: guessing this is a `to: "inventory"` too?

collections-plan.md
```diff
@@ -0,0 +1,102 @@
...
+**start collection states:**
+inventory [{key: "helmet", qty: 1}]
+player []
+
+**Command**
+`collections.moveItem({ key: "helmet", from: "inventory", to: "player", fromState: "equipped")`
+
+**end collection states:**
+inventory [{key: "helmet", qty: 0, state: "equipped"}]
+player [{key: "helmet", qty: 1}]
```
AdamXA, line 93: something feels a bit weird confusing around the inventory or player handling the state of equipped or not? maybe just confuses me with the naming of the collections.

I'm not sure whether a collection called `equipped` and `owned` would solve that? the item goes into a collection for that rather than getting a state in a collection?

collections-plan.md
FostUK: you mean the wording "fromState"? Maybe can change that to something more explicit...

collections-plan.md
FostUK: lol. Yep.!

collections-plan.md
FostUK: Will have a think but that would mean 5 collections and the need to merge them to get the shop list.

BeeGodwin: Yeah, this seemed a bit odd to me to. If I equip a thing, conceptually I still think of it as being in my inventory.

Given that the collections concept doesn't have us create an object per item sold I'm wondering why we don't just have flags and/or quantities on items in collections. I know that's been discussed and dismissed, I am not sure if I understand why though.

BeeGodwin: (It would have the advantage of there just being one list- which seems like it should be nice & simple...)

FostUK: > It would have the advantage of there just being one list- which seems like it should be nice & simple...
A collection is a set of filters that pull from the catalogue and then apply any difference stored in local storage.

The single collection concept unravels somewhat when you start talking about saves because two shops might pull in the same catalogue item. I think we make the mistake of thinkiing of this like an atomic database but it can't be because same item can appear in different collections with different quantities and states. If we actually designed a classic linked table db round that it would probably make more sense but wouldn't necessarily be as easy to serialise. It keeps running through my head that might be the way to go but since the above will work (hopefully) without tearinng apart the gmi storage code I keep mobong in that direction.

If you aren't in some kind of atomic db where every item is tracked you would run into less problems but we are 
With the single store concept it gets a bit wierd when you have multiple shops and 

> Yeah, this seemed a bit odd to me to. If I equip a thing, conceptually I still think of it as being in my inventory.
Actually I think you're both right here. Inventory should just set "equipped" state and we do away with player. 

> I'm not sure whether a collection called equipped and owned would solve that? the item goes into a collection for that rather than getting a state in a collection?

This would make merging a lot more complex. I'm leaning more towards inventory has a state of "equipped" and maybe the shop does the compare of the two for purchased. The problem I had with that is there's no way to know if you started with an object or you actually bought it freom a shop. Also what happens with multiple shops when you buy from one and not the other?

FostUK:
@BeeGodwin / @AdamXA - Updated the above.

**Changes from previous plan:**
* Player collection is gone - intead we just mark state as "equipped" in the inventory collection
* The `moveItem` method is gone. Instead replaced with a call to `set` on each collection. I think this makes more sense as you aren't really moving anything - you're manipulating both collections. 
* Set needs to support +1 and -1 for numeric values. Not quite worked out how yet but I think it will the code shorter instead of having to pull in the current values.


Probably the only thing I'm not 100% on right now is setting shop item "state" to "owned" - might just be the name "state" and label would work better - easy enough to change later.

BeeGodwin (approved):
peachy 🍑

PR #337, opened 2020-11-18
[CGPROD-2914] Fix accessible element position on gel buttons

AdamXA:


FostUK (approved):
Ship it!

🚢🛳️🧊

AdamXA:
pulled the crash fix out of this since it doesn't work, rest is the same

PR #338, opened 2020-11-18
[CGPROD-2768] shop confirm screen - part 1

BeeGodwin:
- Adds layout for the confirm screen
- Resize and visibility functions reused from the shop menu
- Refactor menu-buttons.js for reuse
- Placeholder images etc for items
- Not hooked up to anything yet!


src/components/shop/confirm.js
```diff
@@ -0,0 +1,72 @@
...
+
+export const createConfirm = (scene, config, bounds) => {
+    const { buttonsRight } = config.menu;
+    const { styleDefaults } = config;
+
+    const confirmContainer = scene.add.container();
+    const innerBounds = getOffsetBounds(bounds, getInnerRectBounds(bounds, buttonsRight));
+    const yOffset = bounds.height / 2 + bounds.y;
+
+    confirmContainer.buttons = createConfirmButtons(scene, innerBounds, config, yOffset); // hard coded value nono
```
FostUK, line 19: comment?

src/components/shop/confirm.js
```diff
@@ -0,0 +1,72 @@
...
+
+    confirmContainer.add([
+        createRect(scene, getHalfRectBounds(bounds, !buttonsRight), 0xff0000),
+        createRect(scene, getHalfRectBounds(bounds, buttonsRight), 0xff00ff),
+        createRect(scene, innerBounds, 0x0000ff),
+        scene.add.text(innerBounds.x, promptY(bounds), config.confirm.prompts.buy, styleDefaults).setOrigin(0.5),
+        scene.add.text(innerBounds.x + 20, currencyY(bounds), "1000", styleDefaults).setOrigin(0.5),
+        scene.add.image(innerBounds.x - 20, currencyY(bounds), "shop.balanceIcon"),
```
FostUK, line 27: Possibly using "shop.balanceIcon" might be a bad idea - e.g: there could be multiple shops and they'll be named whatever they are called in `main,js`.

If you need to merge any of this now though just make a note we need to cover these options before release.

FostUK (approved):
Looks good - just some really minor comments :) 

🌭

src/components/shop/confirm.js
BeeGodwin: Yeah, this should be pulling a value from config- fixed.

src/components/shop/confirm.js
BeeGodwin: Good spot! removed.

PR #339, opened 2020-11-19
[CGPROD-2900] Remove extra navigate callback

AdamXA:
some reason there's a callback on both the continue text and the image and they were both firing navigate to the results screen events at the same time.


BeeGodwin (approved):
Lovely! 👍 🌵 🐟 🌸

PR #340, opened 2020-11-19
[CGPROD-2768] shop confirm screen part 2

BeeGodwin:
- confirm.js has an `update` which populates its elements with the appropriate item details, called from the scrolly list buttons
- transact.js added, initially just with the `buy` transaction which sets values on the relevant collections
- routing through the Shop updated with a stacking mechanism on the back button
- User flow now goes shop / manage => pick item => confirm or cancel => transact or not & return to shop / manage.
- Confirming a transaction updates the balance UI (right now balance is just a value pulled from shop config, this will change)
- Added a long description field to items, show in detail view on confirm (as per Figma designs)
- add tests etc.


FostUK (approved):
This look great! 
Just for future consideration- best to try and avoid adding props to Phaser objects. e.g: here you create a container and then add new properties to it. Better to create a parent "confirmation" object with a property called "container" and hang all the other parts off the parent. We're equally guilty of doing the same in various parts of the codebase so it's totally ok as is - just something to bear in mind with any new work :)

PR #341, opened 2020-11-24
[CGPROD-2910] fix select debug crash

BeeGodwin:
- Work to move the Select screens to the new collections system had not encompassed changing the relevant debug configs.
- This left us with crash bugs on debug, as Select expects a collection name in config but our debug examples weren't providing it. Added.
-  Removed choices arrays from both configs as we now use the collections system and this was no longer needed.

### DIFF
--- a/debug/examples/select-1.json5
+++ b/debug/examples/select-1.json5
@@ -1,216 +1,135 @@
 {
-            debugLabels: [
-                {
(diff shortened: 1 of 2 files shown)


FostUK (approved):
Ahh ! I missed this when I coverted collections over. Dur!

PR #342, opened 2020-11-24
[CGPROD-2769] equip confirm screen

BeeGodwin:
- Mostly, just reuses existing functions created for the shop confirm screen
- Adds an equip transaction
- Does not visually update anything (this will be done in CGPROD-2928) but we can verify: dev tools => application => local storage => check state of local storage before & after the transaction.


FostUK (approved):
Looks great! 🚙🚁🚌🚝🚕🚜

PR #343, opened 2020-11-24
[CGPROD-2913] fix resize of scrollable list

BeeGodwin:
- We had some situations where resizing the scrolly list would break the layout
- turns out that as well as a resize subscription, we also needed to add to onScaleChange


FostUK (approved):
Looks great!

PR #344, opened 2020-11-25
[CGPROD-2915-b] Fix duplicate button click in demo game

BeeGodwin:
- Button text was interactive, as was the button graphic itself, resulting in duplicate clicks.
- Removed the `onPointerUp` event from the text elem
- Also removed a redundant `setInteractive`
- Update tests


FostUK (approved):
🦖

PR #345, opened 2020-11-25
[CGPROD-2911] fix crashy click handlers

BeeGodwin:
- Various buttons in the shop were lacking an unsubscribe on shutdown
- This meant that after the first launch of the shop in a given session, the event bus had two callbacks for each click, one of which would have a null pointer and therefore cause a crash.
- Added unsubscribe on shutdown via `scene.events.once`


FostUK (approved):
🦕

PR #348, opened 2020-11-26
[CGPROD-2928] Shop MVP loose ends

BeeGodwin:
- Scrollable list items now reflect the item's state in the collection on initial render
- Items update when their parent list becomes visible- ie if we now own or have equipped an item
- Scrollable list rebuilds itself if the underlying collection has changed- ie if we have different items
- Illegal transactions are now reflected in the UI - button is tinted & an alpha is applied
- Confirm button returns early (noop) on an illegal tx (we are still sending the click event, though.)

### DIFF
--- a/src/core/layout/scrollable-list/scrollable-list-buttons.js
+++ b/src/core/layout/scrollable-list/scrollable-list-buttons.js
@@ -8,6 +8,7 @@ import { handleClickIfVisible } from "./scrollable-list-handlers.js";
 import { eventBus } from "../../event-bus.js";
 import { overlays1Wide } from "./button-overlays.js";
 import { accessibilify } from "../../accessibility/accessibilify.js";
+import { collections } from "../../collections.js";
 import fp from "../../../../lib/lodash/fp/fp.js";
 
 const STATES = ["cta", "actioned"];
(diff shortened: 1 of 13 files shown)


src/core/layout/scrollable-list/scrollable-list-buttons.js
```diff
@@ -64,27 +65,50 @@ const scaleButton = (gelButton, layout, space) => {
     gelButton.setScale(scaleFactor);
 };
 
-const toggle = button => () => {
+const updateButton = button => {
```
FostUK, line 68: Could one line this.

src/core/layout/scrollable-list/scrollable-list-buttons.js
```diff
@@ -64,27 +65,50 @@ const scaleButton = (gelButton, layout, space) => {
...
+        return true;
+    };
+
+    return fp.isEqual(button.item, item) ? false : doUpdate(button, item);
+};
+
+const getState = (item, title) => {
+    const isShop = title => (title === "shop" ? true : false);
+    const isOwned = item => item.state && item.state === "owned";
+    const isEquipped = item => item.state && item.state === "equipped";
```
FostUK, line 87: Can use optional chains now so this could be
item => item?.state === "equipped"

FostUK (approved):
Looks great. Some super minor comments but more just for future ref.

🎄🎄🎄🎄🎅🏿🤶🎄🎄🎄🎄

src/core/layout/scrollable-list/scrollable-list-buttons.js
BeeGodwin: 👍

src/core/layout/scrollable-list/scrollable-list-buttons.js
BeeGodwin: 👍

PR #349, opened 2020-11-26
CGPROD 2930 Resize Issues

FostUK:
* Fix alternating menu on resize

### DIFF
--- a/src/components/shop/balance-ui.js
+++ b/src/components/shop/balance-ui.js
@@ -6,7 +6,8 @@
  * @license Apache-2.0
  */
 
-import { getXPos, getYPos, getScaleFactor } from "./shop-layout.js";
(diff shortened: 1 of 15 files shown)


BeeGodwin (approved):
This looks great. There was another AC in the ticket tho IIRC (basically when you resize & the aspect ratio changes the text gets stretched) ...are we punting on that for now?

FostUK:
> This looks great. There was another AC in the ticket tho IIRC (basically when you resize & the aspect ratio changes the text gets stretched) ...are we punting on that for now?

WIP label still on this :)

Also this fix doesn't actually work due to waht I mentioned earlier. Just fixing the unit tests round this but then will look at the other part.

AdamXA (approved):
doesn't look like a horrible merge for me at least 😂

PR #350, opened 2020-12-01
[CGPROD-2922] Fix accessibility in scrollable list

AdamXA:
accessibilifying the label instead of the gel button works for the position. however, the label does not have input events on it, and the gel button does


BeeGodwin (approved):
Tested locally, works great, lovely stuff, have a hamster 🐹

AdamXA:
I do like hamsters

PR #351, opened 2020-12-02
[CGPROD-2844] use currency from inventory

BeeGodwin:
- relevant code now uses the quantity of a nominated inventory item as currency
- currency item is filtered out of the scrollable list
- transactions update the quantity of that nominated item in the inventory collection
- panel updating checks against the filtered collection

### DIFF
--- a/src/core/layout/scrollable-list/scrollable-list.js
+++ b/src/core/layout/scrollable-list/scrollable-list.js
@@ -11,15 +11,15 @@ import { collections } from "../../collections.js";
 import { onScaleChange } from "../../scaler.js";
 import fp from "../../../../lib/lodash/fp/fp.js";
 
-const createPanel = (scene, title, prepTx) => {
-    const panel = scene.rexUI.add.scrollablePanel(getConfig(scene, title, prepTx));
+const createPanel = (scene, title, prepTx, parent) => {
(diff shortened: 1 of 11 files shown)


src/core/layout/scrollable-list/scrollable-list.js
```diff
@@ -122,16 +123,19 @@ const updatePanelList = panel => {
     const tableContainer = panel.getByName("gridContainer", true);
     const scene = panel.parentContainer.scene;
     tableContainer.clear(true);
-    tableContainer.add(createTable(scene, panel.name, panel.callback));
+    tableContainer.add(createTable(scene, panel.name, panel.callback, panel.parentContainer));
```
FostUK, line 126: Thinking about this - could you just pass panel in? (I realise this isn't exactly part of this exact PR... :)

FostUK (approved):
Looks great - I feel a bit more confident about that idea of using an item for currency now that I can see it in code too :)

src/core/layout/scrollable-list/scrollable-list.js
BeeGodwin: `createTable` gets called at construct time and exists before `panel` does in that context. Suppose I could add a wrapping fn that destructures the panel and calls `createTable` but I don't think it'd be any clearer...

PR #352, opened 2020-12-02
[CGPROD-2844] Allow debug screens to be hidden

BeeGodwin:
- Debug launcher filters out screens from examples.js that have a property `hidden: true`
- These screens are still available to route to, so we can now build more complex flows inside debug.


FostUK (approved):
Great!
🍔

PR #353, opened 2020-12-03
[CGPROD-2844] shop demo game

BeeGodwin:
- Add demo game button to debug launcher which launches sub-menu
- Add hidden screens to debug for demo shop and demo shop game
- Add the game itself with a one-button mechanic; press space to chop (or click, or tap)
- Currency is gettable in game, spendable in shop on hats, which can be equipped (all at once)
- Player spritesheet is selected based on most expensive equipped hat
- Tests are intentionally super minimal and loose - just enough for coverage- can be tightened up, probably not needed
- Add SFX
- Add cheat buttons to add coins, reset economy (crudely, via a gmi call- we will prob want a better mechanism in future.)
- Add touch / click input


debug/config.json5
```diff
@@ -1 +1,4 @@
-{}
+{
+    collection: "hats",
+    // physics: { default: "arcade", arcade: {} }, // why doesn't this work?
```
BeeGodwin, line 3: This needs looking at. Right now the only way I can get arcade physics in debug mode is to drop it into main.js as a global setting.

src/core/debug/examples.js
```diff
@@ -155,11 +156,39 @@ const exampleScreens = {
...
+            shop: "debug-shop-demo-shop",
+        },
+    },
+    "shop-demo-game": {
+        scene: ShopDemoGame,
+        title: "Shop Demo Game",
+        routes: {
+            back: "debug-shop-demo",
+        },
+        // settings: {
```
BeeGodwin, line 174: This is how we usually enable physics per-scene, but doesn't work when the scene is called from debug.

src/core/debug/shop-demo.js
```diff
@@ -21,6 +21,7 @@ class ShopDemo extends Screen {
 
 class ShopDemoGame extends Screen {
     create() {
+        this.plugins.addToScene(this.sys, [], [["ArcadePhysics"]]);
```
BeeGodwin, line 24: ahh nice! I actually tried this but didn't realise that the 3rd arg needed to be an array of an array of a plugin (the docs don't mention that...)

BeeGodwin: awful lot of magic string values in Phaser, huh.

FostUK: Yeah it's not an official way of doing things but I think ok for a debug screen. Not clear what's supposed to be private sometimes.

FostUK (approved):
Looks great - I'm in two minds whether we should folder this and split shop-demo.js into shop demo menu and shop demo game? Just in case anyone ends up copying this as a starting point.
Are we likely to work on this some more for any reason?

BeeGodwin:
Just fired it up to test before merging and it's broken. Investigating...

PR #354, opened 2020-12-11
[CGPROD-2945] shop backgrounds

BeeGodwin:
- Outer padding of scrollable list can be configured as larger / smaller than the inner padding values
- A pane background can be assigned per shop, or per pane, or can default to a transparent bg
- Transparent BG is created with a transparent rect rather than an image
- Switch out development placeholder assets for better ones, hence the bloated PR
- Remove icons from menu buttons; UX may want configurable buttons later, for now- text label on a background.
- Reflow confirm pane to use 2 text objects rather than 3- pending further discussion on IA of items
- Update tests


themes/default/shop/config.json5
```diff
@@ -1,4 +1,5 @@
 {
+    // Needs updating
```
AdamXA, line 2: what needs updating :D

debug/examples/shop.json5
```diff
@@ -49,13 +49,15 @@
     listPadding: {
         x: 18.86,
         y: 15.09,
+        larger: 1.8,
```
AdamXA, line 52: Is this outer or inner padding, what is larger than what?

debug/examples/shop.json5
```diff
@@ -73,13 +75,19 @@
     },
     columns: 1,
     assetKeys: {
-        background: "background",
+        // background: "background", // if a string is passed, that asset is used for all panes
+        background: {
+            menu: "", // leave blank for a transparent BG
```
AdamXA, line 79: if you don't provide a `background.menu` would it error or produce a transparent BG too?

src/components/shop/shop-layout.js
```diff
@@ -79,4 +79,22 @@ export const getInnerRectBounds = scene => {
...
+
+export const getPaneBackgroundKey = (scene, pane) => {
+    const {
+        assetPrefix,
+        assetKeys: { background },
+    } = scene.config;
+
+    if (typeof background === "string")
+        return background ? `${assetPrefix}.${background}` : `${assetPrefix}.noBackground`;
+    return background[pane] ? `${assetPrefix}.${background[pane]}` : `${assetPrefix}.noBackground`;
```
AdamXA, line 99: I'm not really a fan of adding an empty image, to deal with the no background scenario (if that's what it is actually doing). Can we not just check whether a background key is provided before creating a background?

themes/default/shop/config.json5
BeeGodwin: that whole f'in file lol. copypasta fortunately. good spot =)

debug/examples/shop.json5
BeeGodwin: Yeah, that needs to be clearer. TY

debug/examples/shop.json5
BeeGodwin: Omitting the key entirely produces a transparent BG- I've updated the comment for clarity.

themes/default/shop/config.json5
BeeGodwin: a9c902c

debug/examples/shop.json5
BeeGodwin: a9c902c

debug/examples/shop.json5
BeeGodwin: a9c902c

src/components/shop/shop-layout.js
BeeGodwin: Having a play with this. Yes it's definitely not the nicest way to do it. IIRC from Fri something broke if there was no image at all.

BeeGodwin: Uhhhh yeah that one is weird. Basically at one point the menu was an object with a container and a couple of functions hanging off it. The fns use the container bounds to do layout on resize. But at this point everything that was in the container has been refactored out, apart from the background. If there's no background, a bunch of stuff breaks in the menu because it's an empty container and therefore has zero bounds. Could be changed but would need a large-ish refactor and I'm not sure it's worth it...

BeeGodwin: Took a middle path on this: we still have a rect in there giving us an object with non-zero bounds to play with, but it no longer relies on a 16x16 transparent png file, so it's slightly less stinky. 3934321

src/components/shop/menu.js
```diff
@@ -5,23 +5,31 @@
...
-    container.memoisedBounds = bounds; // workaround, pending CGPROD-2887
+    container.memoisedBounds = bounds;
 
-    container.add([
-        createRect(scene, getHalfRectBounds(bounds, !buttonsRight), 0xff0000),
-        createRect(scene, getHalfRectBounds(bounds, buttonsRight), 0xff00ff),
+    const contents = [
         createRect(scene, getInnerRectBounds(scene), 0x0000ff),
-    ]);
+        // createRect(scene, bounds, 0x000000),
```
AdamXA, line 28: comment

src/components/shop/menu.js
```diff
@@ -5,23 +5,31 @@
...
+        createPaneBackground(scene, bounds, "menu"),
```
AdamXA, line 29: so are you saying if there is no key we cannot omit this call? what is relying on it

BeeGodwin: We've got a container that keeps resizing to the safe area. The menu and confirm panes then use that container's bounds to do their layout. The rect on line 27 is basically the button positions. The createPaneBackground call on line 29 has to return some gameObject that's sized to the safe area. (The preferred approach would have been to have the menu and confirm pages get the safe area directly and do layout based on that- however, that fell foul of Genie's debouncing of resize events, which, if I'm remembering correctly, created a race condition.)

BeeGodwin: guess the lesson here is 'ugh containers'

themes/default/shop/config.json5
```diff
@@ -2,7 +2,7 @@
     assetPrefix: "shop",
     music: "home.backgroundMusic",
     background: {
-        items: [{ key: "home.background", x: 0, y: 0 }],
+        items: [{ key: "shop.magicShop", x: 0, y: 0 }],
```
BeeGodwin, line 5: This whole config file is _probably_ orphaned but I've just conformed it to the same as the debug config for now. We might be able to ditch this once upcoming tix around collections and debug mode but at present we need a shop to exist in main.js.

AdamXA (approved):
we will need docs for all this config in future - not sure if there's a ticket to do this

BeeGodwin:
> we will need docs for all this config in future - not sure if there's a ticket to do this

There's not a ticket yet, but Graham is aware- I'll mention it again in post-flight. TY for the 👍

PR #356, opened 2020-12-17
CGPROD-2885_Last-Screens-Music-Shortly-Plays-After-Pausing

Leonie-:
Ensures previous screen's audio does not fade out when returning to the Home screen.

This means that when we pause in-game then return to the home screen, the game music does not fade out before playing the Home screen music.


src/core/game-sound.js
```diff
@@ -23,12 +23,10 @@ const isAlreadyPlaying = audioKey => {
     return audioKey && Assets.backgroundMusic && audioKey === Assets.backgroundMusic.key;
 };
 
-const onFadeComplete = scene => {
-    Assets.backgroundMusic.destroy();
-    startNextMusic(scene);
-};
-
 const startNextMusic = scene => {
+    if (Assets.backgroundMusic) {
```
FostUK, line 27: Can maybe do this now: **Assets.backgroundMusic?.destroy()**

I think that works...

Leonie-: Nice! Will try it.

FostUK (approved):
Looks great.

PR #357, opened 2020-12-17
CGPROD-2907_Adds-Theme-To-Genie

Leonie-:
Allows theme to be passed through as a URL parameter for preference over `embedVars.configPath`.


FostUK (approved):
Looks great. Mo will be super happy about this :)
🐕

PR #359, opened 2021-01-04
[CGPROD-2975] fix a11y clicks on scrollable list

BeeGodwin:
- code in `scrollable-list-handlers` that blocked clicks on non-visible list items also blocked a11y clicks
- added `isA11yClick` fn to infer whether or not the event was fired by a11y element
- it's an a11y click if:
   - we have an activePointer.id of 0 - meaning, we have never had a touch / click event 
   - we have a non zero pointer id, but the most recent upTime is old (arbitrary value of >50ms)

And why yes, that second condition is perhaps a little bit hacky :) open to a better solution if anyone's got one!


Leonie- (approved):
Looks good 🌈

PR #360, opened 2021-01-05
[CGPROD-2981] fix tab order on shop pane navigation 

BeeGodwin:
- Add a11y.reset call on changing panes, update tests.
- Tested with keyboard tabbing and VoiceOver
- some oddness with highlighting on VoiceOver, but this appears to be coming from the rexUI plugin- may not be in our gift to fix. However, the voiceover is consistent with our other pages, and tabbing definitely works nicely.


FostUK (approved):
Looks great! 🎄🌳🌲🌴🎋

PR #362, opened 2021-01-06
CGPROD-2907_Automates-Theme2-for-Genie

Leonie-:
Automates the process of creating `theme2` - copies over the `default` theme then makes all the images greyscale, so you can visually tell the difference between the themes.

The bash script in this PR will also be run in the starter-pack-copy Jenkins job to ensure it is created for the Starter Pack.

Screenshot 2021-01-06 at 10 29 03 am


BeeGodwin (approved):
Looks great, and I ❤️  anything that makes the repo smaller and easier to maintain!

PR #363, opened 2021-01-08
[CGPROD-2966] Make debug screens use their own collections for shop

AdamXA:
the names of some of these example collections and demo screens are a bit confusing, any suggestions welcome 😄


test/core/debug/debug-screens.test.js
```diff
@@ -50,6 +52,7 @@ describe("getDebugScreens", () => {
                 navigation: { "debug-mockKey1": "", "debug-mockKey2": "", "debug-mockKey3": "" },
                 screen: "mockConfig",
             });
+            expect(Load.loadCollections).toHaveBeenCalledWith(mockScreen, mockScreen.context.config, "debug/");
```
AdamXA, line 55: couldn't separate this into another test, something weird going on, probably because the function `addExampleScreens` is a singleton?

babel.config.js
```diff
@@ -10,7 +10,7 @@ module.exports = function (api) {
         [
             "@babel/preset-env",
             {
-                useBuiltIns: "entry",
+                useBuiltIns: "usage",
```
AdamXA, line 13: `async` `await` only worked with this set to usage.

FostUK (approved):
Looks great! 🏕️⛺🥾🚱

test/core/debug/debug-screens.test.js
FostUK: It's debug so prob not worth worrying about too much.

BeeGodwin:
Seems like it might have introduced a bug? If you reset your economy from inside the shop demo game, on collecting the first coin you end up with 1001 coins.

debug/items/debug-demo-inventory.json5
```diff
@@ -0,0 +1,9 @@
+{
+    catalogue: "debug-game-items", //catalogue array or string filename of json5 catalogue
+    include: ["starter-item"], //include all items with this tag in this collection
+    defaultQty: 0, //default quantity unless specified
+    defaults: [
+        { id: "coin", qty: 1000 },
```
BeeGodwin, line 6: o i c, not a bug but a feature? It looks odd tho since the 1000 isn't applied at construct time for some reason.

AdamXA: hmm dunno, is it cause the defaultQty is 0? might be a bug made visible by this 👍

AdamXA: oh if its local storage then fine :D

PR #364, opened 2021-01-08
[CGPROD-2978] fix resize crash after using a screen with a scrollable list

BeeGodwin:
- To get resize working nicely scrollable-list needed to be resized twice- once by the scaler (`onScaleChange.add()`) and once by Phaser (`scene.events.on("resize", [fn])`) - this resolved an issue with RexUI not playing nicely with our scaler.
- We weren't removing the scene.events.on listener, resulting in an NPE on resize once the user had visited the shop.
- Added a shutdown listener that unsubs from the Phaser event, resolving the NPE.
- Add test.


FostUK (approved):
Looks great! 
🦐

PR #365, opened 2021-01-11
[CGPROD-2930] genie shop resize issue

BeeGodwin:
- Fix scaling maths that we were applying to text elements on the confirm pane to preserve aspect ratio
- Add getter to confirm pane for its image and text elements as not all elements were getting scaled
- New scaling maths for image elements, respecting any scaling that was needed on the image.


src/components/shop/confirm.js
```diff
@@ -58,9 +58,15 @@ export const createConfirm = scene => {
     container.setBalance = bal => balance.setText(bal);
     container.getBalance = () => balance.getValue();
     container.setLegal = setLegal(container);
+    container.getTextElems = () =>
+        [container.elems.prompt, container.elems.price, ...container.elems.item].filter(isText);
```
Leonie-, line 62: Why not just return all `container.elems` instead of destructuring specific ones?

Leonie-: Just thinking this might make it a bit more future-proof...

BeeGodwin: There needs to be a bit of destructuring as some of the elems are single text / image elems, but the item is an array of elems (since we can have a couple of different views on the item.) Can simplify it though- doing that.

Leonie- (approved):
Looks great 🍁

PR #366, opened 2021-01-12
CGPROD 3002 Phaser Update

FostUK:
paladin
* Add bbcode plugin.
* Update rexui plugin.
* Use Phaser's version of Spine plugin. 
* Update Phaser to 3.5.0. 
* Prep for release


AdamXA (approved):
nice code

PR #367, opened 2021-01-12
NOTICKET debug collection Fixes

FostUK:
* Fix the pathing issue.
* Also add `debug=nolog` option to remove loader log that only QA use

### DIFF
--- a/src/core/loader/load-collections.js
+++ b/src/core/loader/load-collections.js
@@ -10,7 +10,6 @@ const getKey = item => item.collection;
 const loadToCache = screen => key => screen.load.json5({ key: `items/${key}`, url: `items/${key}.json5` });
 const getKeys = config => Object.values(config).map(getKey).filter(Boolean);
 
-/* istanbul ignore next */
 export const loadCollections = (screen, config) => {
     const keys = getKeys(config).flat();
     keys.forEach(loadToCache(screen));
--- a/src/core/loader/load-collections.js
+++ b/src/core/loader/load-collections.js
@@ -10,6 +10,7 @@ const getKey = item => item.collection;
 const loadToCache = screen => key => screen.load.json5({ key: `items/${key}`, url: `items/${key}.json5` });
 const getKeys = config => Object.values(config).map(getKey).filter(Boolean);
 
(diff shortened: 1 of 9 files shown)


test/core/loader/load-collections.test.js
```diff
@@ -53,24 +53,24 @@ describe("loadCollections", () => {
...
-            expect(mockScreen.load.json5.mock.calls[1][0].key).toEqual("somepath/items/collection2");
+            loadCollections(mockScreen, mockConfig);
+            expect(mockScreen.load.json5.mock.calls[0][0].key).toEqual("items/collection1");
+            expect(mockScreen.load.json5.mock.calls[1][0].key).toEqual("items/collection2");
 
-            expect(mockScreen.load.json5.mock.calls[0][0].url).toEqual("somepath/items/collection1.json5");
-            expect(mockScreen.load.json5.mock.calls[1][0].url).toEqual("somepath/items/collection2.json5");
+            expect(mockScreen.load.json5.mock.calls[0][0].url).toEqual("items/collection1.json5");
+            expect(mockScreen.load.json5.mock.calls[1][0].url).toEqual("items/collection2.json5");
         });
```
AdamXA, line 62: just becomes the same test as above now

src/core/loader/load-collections.js
```diff
@@ -7,25 +7,24 @@ import { initCollection } from "../collections.js";
...
         const cataloguesLoaded = () => {
-            keys.forEach(initCollection(screen, path));
+            keys.forEach(initCollection(screen));
             resolve();
         };
 
         const collectionsLoaded = () => {
             const catalogueKeys = fp
-                .uniq(keys.map(key => screen.cache.json.get(`${path}items/${key}`).catalogue))
+                .uniq(keys.map(key => screen.cache.json.get(`items/${key}`).catalogue))
```
AdamXA, line 25: I guess this is fine actually, if we properly name debug collections so that they're unique to what the game actually uses

FostUK: Yeah that's the only issue - collections should be uniquely named anyway. Maybe we can look at this later and making it more robust.

PR #369, opened 2021-01-14
[CGPPROD-2963] improve shop item information architecture and rationalise config

BeeGodwin:
- DRY style merging with new fn on shop-layout (mostly, to force consistency)
- Update scene.add.text() calls to use new styling fn
- Modify confirm.js to create three text fields in item detail view, rather than 2
- Update shop config to reflect & exemplify changes
- Strip explanatory comments from the demo game shop config, since they were duplicates.

### DIFF
--- a/src/components/shop/balance-ui.js
+++ b/src/components/shop/balance-ui.js
@@ -6,15 +6,9 @@
  * @license Apache-2.0
  */
 
-import { getXPos, getYPos, getScaleFactor, getSafeArea } from "./shop-layout.js";
+import { getXPos, getYPos, getScaleFactor, getSafeArea, textStyle } from "./shop-layout.js";
 import { getMetrics } from "../../core/scaler.js";
 import { collections } from "../../core/collections.js";
-
-const styleDefaults = {
-    fontFamily: "ReithSans",
-    fontSize: "24px",
-    resolution: 4,
-};
 
 const makeElement = makerFns => conf => makerFns[conf.type](conf).setOrigin(0.5);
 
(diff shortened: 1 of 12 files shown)


src/components/shop/balance-ui.js
```diff
@@ -6,16 +6,10 @@
...
+import { getXPos, getYPos, getScaleFactor, getSafeArea, textStyle } from "./shop-layout.js";
 import { getMetrics } from "../../core/scaler.js";
 import { collections } from "../../core/collections.js";
 
-const styleDefaults = {
-    fontFamily: "ReithSans",
-    fontSize: "24px",
-    resolution: 4,
-};
-
```
AdamXA, line 18: I always thought these defaults were here, instead of in the config, so that the screen had a fallback, if nothing were provided in config?

BeeGodwin: We've now pushed the defaults into config, since lots of other stuff depends on there being defaults in there already.

BeeGodwin: if we think it's best to have some hard coded defaults these should probably go live next to the new textStyle fn in shop-layout.js- what do you think?

AdamXA: im a bit confused, is this styling all text in the shop? then other things in the config can override it?

BeeGodwin: 8d67f8b

BeeGodwin: So it being in `balance-ui.js` was a bit of leftover baggage basically. Some stuff had defaults, some stuff didn't, bit of a mess etc. 

The state of play now is that everything uses the same default from config (if set) and then any other text elem spec'd in config can have its own `styles` object which merges with the defaults.

AdamXA: ah ok, I guess its good then, as long as the config is clear to read and its easy to see what each thing should be styled with - without having to work out what gets overridden or w/e

AdamXA (approved):

PR #372, opened 2021-01-18
[CGPROD-3012] guard vs race condition on getTopmostSizer 

BeeGodwin:
- Like other buttons in genie, scrollable list buttons are GEL buttons, and are a synthesis of DOM elements and interactive Phaser elements
- Since the scrollable list builds inner elements first, GEL buttons may be live in the DOM before the scrollable list has finished construction
- We were seeing rare, hard to reproduce crashes on calls to `gelButton.rexContainer.parent.getTopmostSizer()` which were _probably_ caused in this way
- Added a guard if `parent` (i.e. the scrollable list) does not yet exist, and a test to document
- This should be fine as any event hitting this time window ought to be happening before the list is visible to the user (i.e. we can assume it was not an intentional click.)

### DIFF
--- a/src/core/layout/scrollable-list/scrollable-list-handlers.js
+++ b/src/core/layout/scrollable-list/scrollable-list-handlers.js
@@ -9,10 +9,13 @@
 import fp from "../../../../lib/lodash/fp/fp.js";
 
 const handleClickIfVisible = (gelButton, scene, handler) => () => {
+    if (!gelButton.rexContainer.parent) return;
+
     if (isA11yClick(scene)) {
         handler(gelButton);
         return;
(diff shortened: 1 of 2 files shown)


AdamXA (approved):
have you tried making the button use the phaser event emitter vs the event bus?
`gelButton.on("pointerup", () => handleClickIfVisible());`

BeeGodwin:
I keep editing comments I mean to reply to. Oopie.

AdamXA:
oh wow, you can edit my comments! 😂

BeeGodwin:
next level inception 💩  there 😹

PR #373, opened 2021-01-18
Cypress e2e

alexparry-brown:
-Adds Cypress
-Adds ability to run tests headless or via cypress terminal
-Adds e2e tests for genie core, debug shop, debug narrative, debug select screes
-Adds screenshots through debug with padding + safe area
-Overwrites any existing screenshots to avoid spam


FostUK (approved):
Looks great. HAd no idea we could access stats like that. :)

PR #374, opened 2021-01-19
Cgprod 2996 local cypress

alexparry-brown:
-function for selecting url based on environment
-docs updated
-function for running cypress tests locally


AnthonyGoodchild (approved):
cool ⛄

PR #376, opened 2021-01-20
[CGPROD-3016] Disable shop buttons that have scrolled offscreen

AdamXA:


BeeGodwin:
Hey- I see what you've done here & it's waaay simpler & cleaner than my original solution, which is awesome. However firing up the shop I can still click on items that are scrolled off the top or bottom (observed on desktop Chrome / macOS.) I think your code might be bumping into the work Nick did to get a11y elems to play nice with RexUI - `scrollable-list.js:84 const forwardEventsToButton () => {...}` - but not sure without further digging. Do ping if you want to chew it over...

AdamXA:
ye its since I merged in master.... will go through this again

AdamXA:
oh it does work fine here, I think the thresholds are just a bit off

AdamXA:
Ok think I've fixed it now

src/core/layout/scrollable-list/scrollable-list-buttons.js
```diff
@@ -41,15 +39,9 @@ const createGelButton = (scene, item, title, state, prepTx) => {
...
-    const callback = () => prepTx(item, title);
-
-    const clickEvent = eventBus.subscribe({
-        callback: handleClickIfVisible(gelButton, scene, callback),
-        channel: gelConfig.channel,
-        name: id,
-    });
-    scene.events.once("shutdown", clickEvent.unsubscribe);
-
+    const callback = () => gelButton.input.enabled && prepTx(item, title);
```
BeeGodwin, line 42: The simplification of the code here is a good win, but maybe we need a different check to `gelButton.input.enabled`? perhaps check `panel.isInTouching()`?

BeeGodwin (commented):
Love a PR that removes code =) but it's not quite there yet?

src/core/layout/scrollable-list/scrollable-list-buttons.js
AdamXA: The problem with this is that then I have to have two different paths of logic to deal with an accessibility click and an actual click - which I'd rather avoid. I think I've got a way to do this, just going to try it

src/core/layout/scrollable-list/scrollable-list-handlers.js
```diff
@@ -10,27 +10,19 @@ import fp from "../../../../lib/lodash/fp/fp.js";
...
+    return !(
+        itemBounds.lower + item.height < visibleBounds.lower || itemBounds.upper - item.height > visibleBounds.upper
+    );
 };
 
-const isA11yClick = scene =>
-    scene.input.activePointer.id === 0 || scene.sys.time.now - scene.input.activePointer.upTime > 50;
-
-const updatePanelOnScroll = panel => () => getPanelItems(panel).map(item => item.setElementSizeAndPosition());
+const enableOrDisableButton = (panel, item) =>
```
BeeGodwin, line 24: OK this makes lots of sense- we're turning off the buttons when they're scrolled off the top / bottom.

BeeGodwin (commented):
Definitely improved- I can still click the non-visible part of a partially visible button though. Does it need to check the click is within panel bounds as well? Thinking on it, it probably needs to check what it's mousing over as well so we don't get the 'active link' pointing hand on those bits.

BeeGodwin (commented):
Clicks outside the panel don't fire, but clicks that fall inside of the inner border padding do, even if the item is hidden. (I feel like I'm nitpicking now tho lol.) I think that `isInTouching()` can be called on any of the elements of the scrollable list- maybe it's a question of drilling down and calling it on one of the objects further down the tree?

BeeGodwin:
Little bit more investigation on this suggests that my idea isn't so great, since the inner panel is actually larger- i.e. you can be outside the panel, but over a hidden item, and you're still inside the inner panel- it's just masked. So it'd need to check `isInTouching()` on the outer panel but somehow detect if the click is in the border area. Argh!

AdamXA:
> Little bit more investigation on this suggests that my idea isn't so great, since the inner panel is actually larger- i.e. you can be outside the panel, but over a hidden item, and you're still inside the inner panel- it's just masked. So it'd need to check `isInTouching()` on the outer panel but somehow detect if the click is in the border area. Argh!

yea I had a mess with this before, I think it's fine like this - especially on touch devices, it wouldn't be noticeable.
if we can find a way to get it a little better without needing to write much code, then we can always come back and improve this further

BeeGodwin (approved):
Let's roll with that then =) if nothing else, this PR pulls out quite a bit of complexity, so if QA does ask for tweaks, they're gonna be easier (guessing it'd be something like the isInTouching() plus a range check of some kind.)

PR #377, opened 2021-01-20
[CGPROD-3013] enable scrollwheel on scrollable list

BeeGodwin:
- We had click-and-drag scrolling behaviour but had omitted to add mousewheel / touch pad scrolling
- Enabled by adding a listener to phaser's `'wheel'` on setting up events
- Scroll speed is currently hardcoded to 'something sensible' - people might legitimately take a view on that :) it can be moved to config if we prefer.


FostUK (approved):
Looks great!

PR #378, opened 2021-01-21
Cgprod 2995 theme2

alexparry-brown:
-Ability to run theme 1 & theme 2 locally and against deployed test url
-Updated docs
-Changed the way environment variables are called as cypress has its own implementation of this.


FostUK (approved):
Looks great!🥤

PR #380, opened 2021-01-22
[CGPROD-2930] fix confirm some more

BeeGodwin:
- Confirm pane layout was broken if we resized the shop while the pane was hidden
- Changed how confirm pane updates its text elements to be more efficient and less broken
- Fix multiple resizes causing image elements to progressively grow or shrink


FostUK (approved):
Looks great!🦊🐺

PR #381, opened 2021-01-25
CGPROD 3019 Safari Compatible Assets.

FostUK:
* Set audio files to use .mp4 (.wav MIME type is incorrect on Cage which breaks Safari)
* Update ports info in docs
* Fix some Markdown paths that were missing


AdamXA (approved):
guessing the mp4s are AAC encoded

FostUK:
> guessing the mp4s are AAC encoded

I actually didn't check but it seems they were luckily. Well remembered!

PR #382, opened 2021-01-27
[CGPROD-2946] - RC vis changes

BeeGodwin:
Bugfix: 
- We load font CSS up front but get `.woff` or `.woff2` files lazily (our Reith css points at several such files.)
- This meant that text elems that were immediately visible on the shop that did not use the default .woff were rendering in a fallback font, first time only.
- Added a util fn that checks if the font is available and adds a callback to update the element style on font load complete if it's not. 
- _Not_ handling this for all text elems in Genie as you might think, since updating the style is comparatively expensive (involves re-rolling textures and communicating with GPU.) But the fix is portable.
- Rolled this fn into an addText fn on shop-layout and replaced scene.add.text calls with shopLayout.addText.
- Updated tests to suit.

Changes that we actually _planned_ to make: 
- Shop nav buttons now have their own text styles config.
- Visual style updates to config as per ticket
- Config changes replicated to both shop debug screens.

Also: 
- Refactor confirm layout functions, now 72.9% more readable.
- Refactor to clear text element functions out of shop-layout, grouping them with the bugfix above.


src/core/layout/text-utils.js
```diff
@@ -0,0 +1,12 @@
+/**
+ * @copyright BBC 2021
+ * @author BBC Children's D+E
+ * @license Apache-2.0
+ */
+
+export const updateStyleOnFontLoad = textElem => {
+    // ie11?
```
BeeGodwin, line 8: I'm not sure if we need to polyfill this- I can't run local Genie in IE11.

FostUK: Just ran this and it crashes
```
Unable to get property 'check' of undefined or null reference
TypeError: Unable to get property 'check' of undefined or null reference
   at updateStyleOnFontLoad (eval code:10:3)
   at text (eval code:41:5)
   at Anonymous function (eval code:24:5)
   at Anonymous function (eval code:29:7)
   at map (eval code:16:5)
   at createTitles (eval code:46:3)
   at createTitle (eval code:19:3)
   at create (eval code:97:7)
   at create (eval code:4:410431)
   at bootScene (eval code:4:409614)
```

I'm wondering if we should just stub `document.fonts` for IE11 so it doesn't crash?

FostUK (commented):
Looks good. 

2 minor things:

"text-utils.js" - could probably do with renaming or it will end up a dumping ground for random code.

"shop-layout.js" seems to have two chunks of unrelated code in it - might be worth splitting off  the text style parts into their own file.

🐱‍👤🐱‍🐉🐱‍👓🐱‍💻🐱‍🏍🐱‍🚀

src/core/layout/text-utils.js
BeeGodwin: Could go one better and switch to a `setTimeout` solution if `!document.fonts`.

FostUK: I really don't want to add browser specific code for stuff like this if we can avoid it.

Tried updating to latest core-js and they haven't added anything for it.

maybe just do a:

```
if (!document.fonts?.check(textElem.style._font)) {
        document.fonts?.ready.then(() => textElem.style.update());
    }
```

BeeGodwin: Agreed. 5fecd82 has this plus a mini refactor following your two other points. TY!

PR #383, opened 2021-02-03
CGPROD 3000 NinePatch

FostUK:
* Adds NinePatch plugin
* Adds ninepatch config support to scrollable list
* Update core-js polyfills to latest

### DIFF
--- a/src/components/shop/confirm.js
+++ b/src/components/shop/confirm.js
@@ -9,11 +9,11 @@ import {
     setVisible,
     resize,
     getInnerRectBounds,
-    createRect,
     getSafeArea,
     createPaneBackground,
     textStyle,
 } from "./shop-layout.js";
+//import { createPaneBackground } from "./background.js";
 import { createConfirmButtons } from "./menu-buttons.js";
 import { doTransaction } from "./transact.js";
 import { collections } from "../../core/collections.js";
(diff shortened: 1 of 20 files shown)


src/components/shop/confirm.js
```diff
@@ -14,6 +14,7 @@ import {
     createPaneBackground,
 } from "./shop-layout.js";
 import { addText } from "../../core/layout/text-elem.js";
+//import { createPaneBackground } from "./background.js";
```
Leonie-, line 17: Should this line be deleted?

FostUK: Ah yes! Good spot. Thanks :)

src/core/layout/scrollable-list/scrollable-list.js
```diff
@@ -169,6 +164,10 @@ export class ScrollableList extends Phaser.GameObjects.Container {
     constructor(scene, title, callback, filter) {
         super(scene, 0, 0);
         this.collectionFilter = filter;
+
+        const config = scene.config.backgrounds?.[title] ?? null;
+        this.background = createBackground[getType(config)](scene, config);
+        this.title = title; //TODO remove, used during debug
```
FostUK, line 170: Ahh. this needs removing too.

Leonie- (approved):
There are quite a lot of TODOs still in here, but otherwise looks good! :blowfish:

FostUK:
> There are quite a lot of TODOs still in here, but otherwise looks good! 🐡

Yeah they are just things I've noticed we need to address before release. Saldy haven't been on top of the shop work as much as I should have been :)

PR #384, opened 2021-02-04
[CGPROD-2947] shop functions as overlay

BeeGodwin:
- Fix config for backgrounds in demo game shop
- Update `shop/config.json5` in theme to reflect current state of play
- Shop now switches on `isOverlay` in config. If an overlay, has an `overlayBack` rather than a `back` button (from gel-defaults.js.)
- Added exported shopOverlay function intended to be used from wherever to launch Shop as an overlay.
- Added `docs/development/shop-component/adding-shop.md` with guidance on including Shop in main.js and as an overlay.

### DIFF
--- a/src/components/shop/shop.js
+++ b/src/components/shop/shop.js
@@ -15,6 +15,7 @@ import { getSafeArea, getXPos, getYPos, getScaleFactor } from "./shop-layout.js"
 import { eventBus } from "../../core/event-bus.js";
 import { createConfirm } from "./confirm.js";
 import * as a11y from "../../core/accessibility/accessibility-layer.js";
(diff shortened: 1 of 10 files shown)


src/components/shop/shop.js
```diff
@@ -110,3 +112,9 @@ export class Shop extends Screen {
         );
     }
 }
+
+export const shopOverlay = (screen, shopKey) => {
```
FostUK, line 116: Interesting adding the stats export - should save agencies having to look it up. 

Possibly should be verb here rather than noun? "addShopOverlay" ? "openShop"?

FostUK (approved):
Looks great. Added one minor comment.

☄️

PR #386, opened 2021-02-09
[CGPROD-2949] consumables in inventory

BeeGodwin:
- Refactor overlay config; remove `isDynamic` in favour of using string templates for text content and asset keys. Remove redundant tests that targeted `isDynamic`.
- Refactor out use of assetPrefix in scrollable list button overlays (just title is using it now)


BeeGodwin:
PR is not updating, closing and raising a fresh one.

PR #388, opened 2021-02-09
[CGPROD-2947] part 2 - add overlay shop to main.js

BeeGodwin:
- Adds a button to game.js that launches the shop as an overlay

### DIFF
--- a/test/game.test.js
+++ b/test/game.test.js
@@ -471,9 +471,8 @@ describe("Game", () => {
         describe("shop button", () => {
             shop.launchShopOverlay = jest.fn();
 
-            beforeEach(() => game.create());
-
             test("launches the shop as an overlay", () => {
+                game.create();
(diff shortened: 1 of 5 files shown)


test/game.test.js
```diff
@@ -467,5 +468,16 @@ describe("Game", () => {
                 expect(accessibilify).toHaveBeenCalledWith({ config: { id: 4, ariaLabel: "Continue" } });
             });
         });
+        describe("shop button", () => {
+            shop.launchShopOverlay = jest.fn();
+
+            beforeEach(() => game.create());
```
FostUK, line 474: Don't know if I would put anything in a beforeEach until there is more than one test unless I was expecting more?

FostUK (approved):
Looks great! 🍪

test/game.test.js
BeeGodwin: Good point! Will tweak.

PR #391, opened 2021-02-12
[CGPROD-3078] resize confirm again, plus loose ends

BeeGodwin:
- Update config of overlay Shop to resolve various issues
- Refactor shop-layout functions to stop using container decoration anti-pattern
- Refactor several shop-layout fns out entirely
- Refactor container decoration out of menu.js
- Refactor container decoration out of confirm.js (in both cases, by making the container a sibling of the fns that were formerly decorations)
- container and menu both now resize using a "destroy / create", rather than "mutate", approach.
- refactor out Shop class fns to do with confirm pane resize / destroy, and refine approach used.
- update tests to reflect design changes
- find and fix race condition which was messing up button text aspect ratio and throwing warnings
- refactor out mocks and tests made redundant by other refactoring
- refactor the refactoring, then refactor the refactored tests. Whew! But now it's shiny. (thanks Adam)


src/components/shop/confirm.js
```diff
@@ -58,21 +69,36 @@ const addConfirmButtons = (scene, container, innerBounds, title, action, item) =
...
     const innerBounds = getOffsetBounds(bounds, getInnerRectBounds(scene));
     const yOffset = bounds.height / 2 + bounds.y;
     container.setY(yOffset);
     createElems(scene, container, getPromptText({ scene, action, item }), item, innerBounds, bounds);
     action === "buy" && itemIsInStock(scene, item) && createBuyElems(scene, container, item, innerBounds, bounds);
-    addConfirmButtons(scene, container, innerBounds, title, action, item);
-    return container;
+    confirmPane.buttons = addConfirmButtons(scene, container, innerBounds, title, action, item);
+    confirmPane.container = container;
+    confirmPane.destroy = destroyContainer(container);
```
AdamXA, line 101: try to avoid overriding functions on the pane, this can be hard to keep track of.
if I'm in another file and I see `pane.destroy()`, since its not an object we've defined, it can be confusing to read as it's usually assumed this function came from the library.

a pattern we usually use instead is to return an object containing all the methods you need - eg. like layout.js, collections.js
if you use this pattern instead, when you goto layout.js, etc, you can see all the functions on this object more clearly - instead of any file being able to manipulate injected objects

AdamXA: I misread this, it has been changed to be like this, which is better, but I was confused by the confirmPane not being a rexUI pane anymore :D maybe we can attach some of these methods like resize, destroy, etc to the confirmPane initially, or create the object at the end, when we return everything (like layout/collections does, etc)

BeeGodwin: 91d6983 hopefully makes things a bit clearer. Wasn't possible to do a completely 'flat' composition at the end of the fn, since the object needs to exist before its `resize` can be created. But now it's a bit more explicit that we're composing a thing, not mutating a thing.

BeeGodwin: bd38887 wrapping it up by updating tests. Nice changes on that pairing call, TY =)

PR #393, opened 2021-02-16
[CGPROD-2952] locked items in Shop

BeeGodwin:
- "locked" state can be set in collection or item list, along with hint text (could do already: added example)
- States and state properties can be described in shop config
- Items can be unlocked by setting the state string on the item in collections (could do already: added example)
- Locked items don't fire their click action
- Properties (tint and alpha) can be configured per item state (i.e. 'locked') and are applied to the button base layer and any flagged overlays
- A 'disabled' member can be added to a state in config to disable buttons that are in that state
- Refactor out a lot of redundant / obsolete test code


src/components/shop/scrollable-list/scrollable-list-buttons.js
```diff
@@ -71,11 +74,13 @@ const getButtonState = (scene, item, title) => {
...
     states.push(isItemInStock(item) ? "available" : "unavailable");
+    states.push(isItemLocked(item) ? "locked" : "unlocked");
     return states;
 };
 
-const isItemEquippable = item => item.slot;
-const isItemInStock = item => item?.qty > 0;
+const isItemEquippable = item => Boolean(item.slot);
+const isItemInStock = item => Boolean(item?.qty > 0);
+const isItemLocked = item => Boolean(item.state === "locked");
```
AdamXA, line 83: don't lock this to a name of a state, instead check whether the button should be enabled/disabled in that state

eg. in select screen we do something like this, where enabled: true/false determines whether the button should be active or not. this allows the state to have any name and allows you to define multiple locked states.

I guess then your activeState would be - enabled/disabled

```
    states: {
        locked: {
            enabled: false,
        }
}
```

BeeGodwin: 1e1df88 re getting rid of the hardcoded 'locked'. Re overlap between states / activeStates, states is a list of state labels on the scrollable list button, whereas activeStates is a list of state labels to use as a filter ('if all these state labels are on the button, add this overlay.') Probably needs a rename, I'll have a think.

BeeGodwin: Ok- in 3a799be, activeStates is now activeInStates, which is hopefully clearer.

PR #394, opened 2021-02-16
[CGPROD-3043] Fixes web audio on safari being stuck in the interrupted state.

AdamXA:
Fixes audio in Safari after using a prompt in the debug screen.
This may also fix a previous iOS bug that I remember. That one prevented audio from playing after switching safari tabs or changing app to a different app in iOS.


FostUK (approved):
Looks great!
🎧

PR #395, opened 2021-02-18
Cypress reports

alexparry-brown:
-updates genie build file, runs cypress against dev webpack server on CI
-sends screenshots to a zip file in output and uploads to S3
-sends test report to output file and uploads to s3
-runs against theme 1 and theme 2
-removes usage of chrome
-sets flag for theme
-generates report


build-scripts/build.sh
```diff
@@ -1,10 +1,59 @@
...
+                node cypress/support/createReports.js
+                zip -r output/reports/screenshots.zip cypress/screenshots
+                $UPLOADTOS3 nolatest
+                exit 1
+        fi  
+}
+
+set -e 
+source /etc/profile
+nvm install 12.19.0
```
AdamXA, line 39: we do this in Jenkins too, is there a reason its also in here

build-scripts/build.sh
```diff
@@ -1,10 +1,59 @@
...
+     if [ $? -eq 0 ]
+            then
+                node cypress/support/createReports.js
+                zip -r output/reports/screenshots.zip cypress/screenshots
+                exit 0
+            else
+                echo "The Cypress tests failed - Uploading report to S3 with no latest" >&2
+                node cypress/support/createReports.js
+                zip -r output/reports/screenshots.zip cypress/screenshots
+                $UPLOADTOS3 nolatest
```
AdamXA, line 14: the way this is wrote, I think, will only work if it's the 4th argument.
https://github.com/bbc/childrens-games-jenkins/commit/9e7482bc33acfcfc6334f078b4c786a5094b4744#diff-3c062d7265d94ac4bf580860782fdfa3e8f2a9e256645c9e5b17cfc629013e9a

i'm not really a fan of having this in the s3upload script though, this is kind of bespoke. might be better separated into another script

AdamXA: I also think this will fail because it will upload to s3, then at the end of the job it will try to upload the built game, but the folder will already exist so it will fail.

cypress/support/appendToken.js
```diff
@@ -0,0 +1,8 @@
+/**
+ * @copyright BBC 2018
+ * @author BBC Children's D+E
+ * @license Apache-2.0
+ */
+import notp from "notp";
+const genToken = () => notp.totp.gen("sdK3uk]'mDCj,,zq");
```
AdamXA, line 7: guess this is fine

build-scripts/build.sh
alexparry-brown: Because its in the shell script, line 38 is needed to it has access to the S3 upload file. 
The version of npm was also getting reset as a result of line 38 back to the default version Jenkins used so we needed to set it again to make sure it uses the correct version

build-scripts/build.sh
alexparry-brown: Ah, we've changed that script again. @AnthonyGoodchild modified it, not sure where you could see those changes. Ant have you got a link out to it?

alexparry-brown: This was the revised script https://github.com/bbc/childrens-games-jenkins/pull/34/files

AnthonyGoodchild: yeah it's just not merged yet as I was reworking how the s3 upload script is called - it's got messy with all the arguments

AnthonyGoodchild: (the command here will currently work as the script is modified on live)

PR #397, opened 2021-02-23
GGPROD-3073 Added routes for shop navigation

AnthonyGoodchild:
fixes pause bug when checking for `restart` and `select` routes.


alex-s2 (approved):
👍

PR #402, opened 2021-02-26
Temp cypress

alexparry-brown:
-reduced the timeouts (quicker time to fail)
-changed logic so it only runs theme 2 against master


build-scripts/build.sh
```diff
@@ -52,13 +52,24 @@ cp -r themes output/themes
...
-#    then
-#        npm run cy:local-theme2
-#        postTheme2Actions
-#    else
-#        npm run cy:local-theme2
-#        postTheme1Actions
-#fi
-
+npm run start:pack & npm run cy:local
+if [ $? -eq 0 ]
```
FostUK, line 56: Can shorten a tiny amount to:

```
if [ $? -eq 0 ]
then
    if [ "$GIT_BRANCH" = "master" ]
    then 
        npm run cy:local-theme2
        postTheme2Actions
    else
        node cypress/support/createReports.js
        zip -r output/reports/screenshots.zip cypress/screenshots
        exit 0
    fi  
else
    if [ "$GIT_BRANCH" = "master" ]
    then 
        npm run cy:local-theme2
    fi  
    postTheme1Actions
fi
```

FostUK (approved):
Looks great - one minor comment.

PR #404, opened 2021-03-03
NOTICKET DOM Elements

FostUK:
Allows DOM elements to show in the correct place.


AnthonyGoodchild:
This looks good (and works) but in chrome device mode the touch-scroll behaviour is back where it will fire click on scrolling by touch.

AnthonyGoodchild:
> This looks good (and works) but in chrome device mode the touch-scroll behaviour is back where it will fire click on scrolling by touch.

looks like this was brought back in an earlier commit, so ignore on this PR

AnthonyGoodchild (approved):
👍😏👍

PR #405, opened 2021-03-03
Cgprod cgprod 3079 consumables

alex-s2:


src/components/shop/transact.js
```diff
@@ -51,6 +52,11 @@ export const use = (scene, item) => {
     const invItem = invCol.get(item.id);
     const qtyLeft = invItem.qty - 1;
     gmi.sendStatsEvent("use", "click", { id: item.id, qty: qtyLeft });
+    eventBus.publish({
+        channel: "shop",
+        name: "used",
+        data: invItem,
```
AdamXA, line 58: how does the game know how many you have used

AdamXA: I guess you get an event each time you use one?

PR #406, opened 2021-03-04
CGPROD-3095 Scrollable-list button disabled state 

AnthonyGoodchild:
Set button.off for `disabled` button states, determined by `disabled` prop
in the shop button config.

This stops audio being played for locked buttons, or any button state which has `disabled` in the shop config.


FostUK (approved):
Looks great!

PR #414, opened 2021-03-07
CGPROD 3082 audio

FostUK:
* Add default audio for buy/equip/unequip/use
* Add ability to configure per item overrides.


AdamXA (commented):
how easy would it be to pass a different button click sound to the gel button?

FostUK:
> how easy would it be to pass a different button click sound to the gel button?

There's no support for it at the moment - it's hard coded. Annoyingly the button config is unavailable at that point too.

Since we need to disable that sound anyway I might look at moving this code to assign a sound rather than play one but there's a spearate ticket to remove it so might as well be done in that.

PR #415, opened 2021-03-08
[CGPROD-2891] Titles v2

AdamXA:


src/components/shop/shop-list-screen.js
```diff
@@ -48,7 +49,8 @@ export class ShopList extends Screen {
 
     resize() {
         this.scrollableList.reset();
-        this.balance.resize();
+        this.titles.title.resize();
```
FostUK, line 52: Could we bury this in scene and then it would be automatic for every screen?
Does it have it's own onscale hook?

AdamXA: can do that 👍

FostUK: Also then you wouldn't need to assign it to the class in the contructor for most applications.

AdamXA:
made changes

PR #417, opened 2021-03-08
[CGPROD-3075] Phaser and RexUI upgrade

AdamXA:


FostUK:
@AdamXA - I updated the Phaser plugin and the files work as is. Then noticed an npm audit issue with webpack. Got version 5 which broke everything. Looks working now - if you are ok with this just merge it. Need to do a draft release to check Stories module. If there are further problems we can fix in another commit maybe?

FostUK (approved):
_I have crossed oceans of time to find this pull request_

AdamXA:
CI check passed on rerun

PR #418, opened 2021-03-08
NOTICKET constructor.name mangling

FostUK:
* Fixes problem with webpack mangling constructor.name (should not rely on constructor.name and use constructor instead)
* Fix demo game tests that were relying on a class property collision between .list (now scrollableList) and Phaser's .list


src/components/shop/shop-menu-screen.js
```diff
@@ -10,10 +10,12 @@ import { onScaleChange } from "../../core/scaler.js";
 import { createTitles } from "../../core/titles.js";
 import { createBalance } from "./balance-ui.js";
 import { createMenu } from "./menu.js";
+import { initResizers } from "./backgrounds.js";
 
 export class ShopMenu extends Screen {
     preload() {
         this.plugins.installScenePlugin("rexUI", RexUIPlugin, "rexUI", this, true);
+        initResizers();
```
AnthonyGoodchild, line 18: can you init the resizers as part of `addBackgroundItems`, or would that be polluting it with Rex?

AnthonyGoodchild (approved):
good fix 👍

src/components/shop/shop-list-screen.js
```diff
@@ -11,18 +11,15 @@ import { onScaleChange } from "../../core/scaler.js";
...
         this.plugins.installScenePlugin("rexUI", RexUIPlugin, "rexUI", this, true);
+        initResizers();
     }
 
     create() {
-        if (this.transientData.shop.mode === "shop") {
-            gmi.sendStatsEvent("shopbuy", "click", {});
-        } else {
-            gmi.sendStatsEvent("shopmanage", "click", {});
-        }
```
AdamXA, line 25: we need this stat right?

AdamXA: hm did this not fail a unit test either

FostUK: That's kind of strange. I never touched that. Will revert and check.

FostUK: Added back in but made it a one-liner.

src/components/shop/shop-menu-screen.js
FostUK: Sadly that is called for every scene - not just ones with the ui plugin included.

Issue #410, opened 2021-03-05
Rename items


FostUK:
The "items" folder in themes should be renamed to "collections" to match config naming.

PR #422, opened 2021-03-09
NOTICKET rename items

FostUK:
tumblr_nu3zivpR0I1qd4q8ao1_500
* Renames /items to /collections in themes
* Addresses: https://github.com/bbc/genie/issues/410


docs/development/select-screen.md
```diff
@@ -77,7 +77,7 @@ Configured choices may have text labels drawn with them. These are populated by
 
 ### Choice states
 
-Choices can have states, see [states](states.md).
```
AnthonyGoodchild, line 80: Might be less confusing to have the states separate and link both there, same if more things have states going forward

FostUK: Yeah. I only made an initial stab at the docs. We still have to update them properly.

PR #425, opened 2021-03-10
CGPROD-3081 Scrollable List Empty State

AnthonyGoodchild:
Adds text to the shop listing when the collection is empty.

This text is configurable through shop config.

Screenshot 2021-03-10 at 15 58 19


src/components/shop/scrollable-list/scrollable-list.js
```diff
@@ -91,7 +86,7 @@ const setupEvents = (scene, panel) => {
     items.forEach(item => getFirstElement(item).addEventListener("focus", () => panel.updateOnFocus(item)));
 };
 
-const getPanelItems = panel => panel.getByName("grid", true)?.getElement("items") ?? [];
```
AnthonyGoodchild, line 94: this will never be empty now, due to the collection size conditional

debug/examples/shop-consumables-list.json5
```diff
@@ -157,4 +157,16 @@
             ],
         },
     },
+    emptyList: {
+        shop: {
+            value: "There is nothing to buy",
+            position: { offsetX: 0, offsetY: 0 },
```
AdamXA, line 163: do we need offsets

AnthonyGoodchild: yes

AdamXA: are these centred anyway?

AnthonyGoodchild: `0, 0` is centred. The config here is for example/show the options.

FostUK (approved):
Ship it

PR #426, opened 2021-03-10
CGPROD 3117 debug confirm backgrounds

FostUK:
* Fix backgrounds on debug screens
* Fix debug confirm screen audio


debug/examples/shop-consumables-confirm.json5
```diff
@@ -30,12 +30,18 @@
         align: "center",
     },
     confirm: {
+        audio: {
+            buy: "buy",
+            equip: "equip",
+            unequip: "unequip",
+            use: "use"
+        },
```
AdamXA, line 38: do we need this as config or can we just reserve these keys for audio?

AdamXA (approved):

debug/examples/shop-consumables-confirm.json5
FostUK: Prob best to let them configure the key since they might not be using the shop and then config this anyway?

PR #429, opened 2021-03-11
3108 timer

alex-s2:


themes/default/collections/game-items.json5
```diff
@@ -19,6 +19,7 @@
         title: "Skull Bomb",
         description: "It'll blow your mind",
         price: 50,
+        duration: 5,
```
AdamXA, line 22: I think shop items have this object you can put pieces of data into:
```
        userData: {
            //.. keep this namespaced out in case agencies want to put object stats in here
        },
```

alex-s2: Duration is meant to be a core thing we support so not sure if it should be here or in there? 

Hmm

I guess since the data can be anything this is fine, an agency could put it in there also, since the game handles what it does with the item it gets back

alex-s2: Will go with this for now and think about it in the config tidy ticket

PR #432, opened 2021-03-11
flag for environment params

alexparry-brown:
-add env variables with flags
-remove mapping of process env values to cypress env variables
-updated docs


FostUK (approved):
top!

PR #435, opened 2021-03-16
CGPROD-3129 getRect

AnthonyGoodchild:
Adds resize for shop menu based on re-calculated bounds.

When resizing the window, the buttons will now better recalculate their correct
size when the bounds have significantly changed.


FostUK (approved):
Looks great! Glad it's sorted :)

PR #436, opened 2021-03-16
CGPROD-3128 Fix showing shop confirm on key press

AnthonyGoodchild:
Key press was not being recognised when no pointer events had previously been fired, added condition for `keydown` on event to call `showConfirmation`.

- Fixes using the return/space key to navigate to shop confirm screen when only tabbing.

- Fixes some unit tests that were not testing anything.


FostUK (approved):
Nice!

PR #438, opened 2021-03-17
CGPROD 3121 confirm backgrounds

FostUK:
* Refactors
* Added icon Background
* Set panel background to fixed image
* Fix aspect ratio
* Lock height


src/components/game.js
```diff
@@ -15,7 +15,7 @@ export class Game extends Screen {
         console.log("Item used:", data); // eslint-disable-line no-console
         console.log("Item timer duration in seconds: ", data.duration || "No duration"); // eslint-disable-line no-console
         let time = data.duration * 1000;
-        let timedEvent = this.time.delayedCall(time, this.onEvent, [], this);
+        this.time.delayedCall(time, this.onEvent, [], this);
```
AnthonyGoodchild, line 18: odd how this wasn't linted

src/components/shop/confirm/confirm-resize.js
```diff
@@ -3,8 +3,7 @@
  * @author BBC Children's D+E
  * @license Apache-2.0
  */
-import { getInnerRectBounds, getSafeArea } from "../shop-layout.js";
```
AnthonyGoodchild, line 6: I wonder if we should remove `getSafeArea` from `shop-layout.js`, should always be able to use `scene.layout.getSafeArea`

FostUK: Yeah I've been doing that where I see it. Just obscures what's going on.

src/components/shop/shop-menu-screen.js
```diff
@@ -26,28 +26,24 @@ export class ShopMenu extends Screen {
         this.transientData.shop = { config: this.config.shopConfig };
         setBalance(this);
 
-        this.menu = createMenu(this);
+        const resize = createMenu(this);
```
AnthonyGoodchild, line 29: I don't like either way of doing this, it reads strange. Not sure how better to do it though..

src/main.js
```diff
@@ -53,7 +53,7 @@ const screens = {
             debug: "debug",
             //Example of custom routing function
             next: scene => {
-                scene.navigate("narrative");
+                scene.navigate("shop-menu");
```
AnthonyGoodchild, line 56: 😂

src/components/game.js
FostUK: Shows up on mine but as a warning.

src/components/shop/shop-menu-screen.js
FostUK: Yeah - might talk about that in the dev channel - it actually seems to happen a lot - twice in this project.

I agree - feels dirty.

src/main.js
FostUK: oh!!! :) Nice spot ta!

src/components/game.js
AnthonyGoodchild: ah, yeah

AnthonyGoodchild (approved):
super fantastic ✌️

src/main.js
AnthonyGoodchild: I keep doing this

PR #442, opened 2021-03-19
NOTICKET Debug Assets

FostUK:
Fix asset configs for debug screens to match main flow


debug/examples/shop-equippables-confirm.json5
```diff
@@ -38,10 +38,12 @@
         },
         buttons: {
             buttonsRight: true,
-            key: "debug-shop.menuButtonBackground",
+            key: "debug-shop.button",
+            cancelKey: "debug-shop.buttonCancel",
             styles: { fontStyle: "bold", fontSize: "16px" },
         },
-        background: { key: "halfViewBackground", columns: [100, null, 100], rows: [100, null, 100] },
+        background: "halfViewBackground",
```
AdamXA, line 45: when should it be asset keys without prefixes and when should it have prefixes?

AnthonyGoodchild: > when should it be asset keys without prefixes and when should it have prefixes?

Aren't you on holiday?

AdamXA: Not today

FostUK: Should always avoid needing the scene key whenever possible. I think sometimes we have added the scene key so it can go in debug but there are ways round that with the asset loader

AdamXA: probably needs some tidy up then so they all follow the same way

PR #456, opened 2021-04-07
CGPROD 3110 button sounds

FostUK:
* Refactor Game sound to music only
* Add ability to set click sound on gel buttons
* Remove but play form transact
* Add sound defaults to menu buttons


src/core/layout/gel-button.js
```diff
@@ -114,6 +109,11 @@ export class GelButton extends Phaser.GameObjects.Container {
         );
     }
 
+    setClickSound(key) {
+        this.config.clickSound = key;
+        this._click = this.scene.sound.add(key);
```
AdamXA, line 114: is there a test for this being called and `play()` being called on it on pause event?

FostUK: Oh good point - there's marginal testing through the contructor but since we can call this directly now I'll add some one ta :)

PR #460, opened 2021-04-09
CGPROD-3169 Add covers to shop-item-list

AnthonyGoodchild:
Adds top and bottom covers to the item lists to mask the scrolling into
the background.

Covers are configured in shop-list config, under the `listCovers` key, e.g:

```
{    
    listCovers: {
        shop: {
            top: { key: "shop-list.listCoverTop" },
            bottom: { key: "shop-list.listCoverBottom" },
        },
        manage: {
            top: { key: "shop-list.listCoverTop" },
            bottom: { key: "shop-list.listCoverBottom" },
        },      
    },
}
```
https://jira.dev.bbc.co.uk/browse/CGPROD-3169

Screenshot 2021-04-09 at 13 34 08


src/components/shop/scrollable-list/scrollable-list-covers.js
```diff
@@ -0,0 +1,55 @@
...
+        top: covers.top
+            ? {
+                  y: safeArea.y - safeArea.height / 2 + padding.y,
+              }
+            : null,
+        bottom: covers.bottom
+            ? {
+                  y: safeArea.y + safeArea.height / 2 - padding.y,
+              }
+            : null,
```
AdamXA, line 22: no need for ternary - if no cover, value not read?
then don't need to pass in `covers` either

src/components/shop/scrollable-list/scrollable-list-covers.js
```diff
@@ -0,0 +1,55 @@
...
+export const resizeCovers = (scene, panel, covers, config) => {
+    if (panel && covers) {
+        const padding = getCoverPadding(config);
+        const positions = getCoverPositions(scene, covers, padding);
+        const panelWidth = panel.getChildrenWidth();
+        covers.top &&
+            covers.top
+                .setX(0 - padding.x)
+                .setY(positions.top.y)
+                .setScale(panelWidth / covers.top.width);
```
AdamXA, line 39: optional chaining might be nicer eg. `covers.top?.setX(...)`

src/components/shop/scrollable-list/scrollable-list-covers.js
```diff
@@ -0,0 +1,55 @@
...
+            : null,
+    };
+};
+
+const getCoverPadding = config => {
+    return { y: config.y * config.outerPadFactor, x: config.x * config.outerPadFactor };
+};
+
+export const resizeCovers = (scene, panel, covers, config) => {
+    if (panel && covers) {
```
AdamXA, line 31: could avoid big if block with a `if (!panel || !covers) return;`

src/components/shop/scrollable-list/scrollable-list-covers.js
```diff
@@ -0,0 +1,55 @@
...
+};
+
+export const createCovers = (scene, config) => {
+    if (scene.config.listCovers) {
+        const top = config.top ? scene.add.image(0, 0, config.top.key) : null;
+        const bottom = config.bottom ? scene.add.image(0, 0, config.bottom.key) : null;
+        return { top, bottom };
+    }
+    return false;
+};
```
AdamXA, line 55: probably can make this a one liner, something like this?
```export const createCovers = (scene, config) => scene.config.listCovers && { top: config.top && scene.add.image(0, 0, config.top.key), bottom: config.bottom && scene.add.image(0, 0, config.bottom.key) }```

having scene.config and config bit confusing to read

AdamXA (commented):
look good 👍

src/components/shop/scrollable-list/scrollable-list-covers.js
AnthonyGoodchild: it's so you can have top without bottom, or the opposite.

src/components/shop/scrollable-list/scrollable-list-covers.js
AnthonyGoodchild: nice didn't think of that :)

src/components/shop/scrollable-list/scrollable-list-covers.js
AnthonyGoodchild: could do, I'm not sure which reads better

src/components/shop/scrollable-list/scrollable-list-covers.js
AnthonyGoodchild: I was considering renaming `config` to `coversConfig`

AnthonyGoodchild: I originally wrote it as returning the object but it didn't seem as clear.

AnthonyGoodchild: I was making a compromise but then ended up writing the same one-liner. I've put it over multiple lines as it's clearer

PR #461, opened 2021-04-09
Shop stats

alex-s2:
```What
CURRENT
Main page = keepalive.games.genie.shop_menu.page

List of items to buy = keepalive.games.genie.shop_list.page

Click on an item to buy = keepalive.games.genie.shop_confirm.page

List of items to manage = keepalive.games.genie.shop_list.page

Click on an item to manage= no idea because I've run out of currency

WHAT WE WANT
All page counternames need minor changes
Main page = keepalive.games.genie.<shop>menu.page

List of items to buy = keepalive.games.genie.<shop>buy.page

Click on an item to buy = keepalive.games.genie.<shop>buyconfirm.page

List of items to manage = keepalive.games.genie.<shop>manage.page

Click on an item to manage= keepalive.games.genie<shop>manageconfirm.page

<shop> being the customizable shop name```

### CODE REVIEW

File: src/core/screen.js [hunk 1]
```diff
@@ -131,7 +134,11 @@ export class Screen extends Phaser.Scene {
         this._layout.makeAccessible();
         this.sys.accessibleButtons.forEach(button => a11y.addButton(button));
         a11y.reset();
-        this.setStatsScreen(this.scene.key);
+        if (this.scene.key.includes("shop")) {
+            this.setStatsScreen(this.transientData.shopTitle + "menu");
+        } else {
+            this.setStatsScreen(this.scene.key);
+        }
```
AdamXA, line 141: `this.setStatsScreen(this.scene.key.includes("shop") ? this.transientData.shopTitle + "menu" : this.scene.key`
probably can do in 1 line

src/core/screen.js
```diff
@@ -81,6 +81,9 @@ export class Screen extends Phaser.Scene {
         data.config && this.events.once("create", nextPage(this));
 
         if (this.scene.key !== "loader" && this.scene.key !== "boot") {
+            if (!this.scene.key.includes("shop")) {
+                this.setStatsScreen(this.scene.key);
+            }
```
AdamXA, line 86: not a big fan of basing these things off scene key but I guess it's the easiest way.
you could prob make this into one big if

src/components/shop/shop-list-screen.js
```diff
@@ -20,6 +20,11 @@ export class ShopList extends Screen {
     }
 
     create() {
+        if (this.transientData.shop.mode == "shop") {
+            this.setStatsScreen(this.transientData.shopTitle + "buy");
+        } else {
+            this.setStatsScreen(this.transientData.shopTitle + "manage");
+        }
```
AdamXA, line 27: `this.setStatsScreen(this.transientData.shopTitle + (this.transientData.shop.mode === "shop" ? "buy" : "manage"));`

can prob make this one line too

src/components/shop/shop-menu-screen.js
```diff
@@ -19,6 +19,8 @@ export class ShopMenu extends Screen {
     }
 
     create() {
+        this.transientData.shopTitle = this.config.title?.text.toLowerCase() || "shop";
```
AdamXA, line 22: I'm a bit confused by this, title could be set to anything, are you after scene.key instead?

alex-s2: yeah wasn't 100% sure what to pick for this, I did deliberately choose this instead of the scene key but who knows

AdamXA: Maybe you need to take the start of the scene key eg. "someshop-menu" becomes "someshop"

alex-s2: sounds good

src/core/screen.js
AnthonyGoodchild, line 86: 😢

AnthonyGoodchild: what if the scene was called "stashopen" or something

src/core/screen.js
AnthonyGoodchild: `if` seems fine here to me, I don't think it reads any better as a one liner

src/core/screen.js
FostUK: Or DROP TABLE GAMES; 

?

FostUK: Yeah I think scene key is maybe the best way handle multiple scenes.

Worth putting anything we might want to refactor at a later date into an issue on github.

FostUK: !["loader", "boot"].includes(this.scene.key)

src/core/screen.js
FostUK: I do tend to strive for less code / lines though as it helps with readability of the whole file.
I've possibly gone a bit too far down the"if staements are a code smell" lately though :)

FostUK (approved):
I think tbh you came up with a decent way to do it with mode - should mean it's independent of the scene key. Def if you think of anything might be worth revisiting stick it in an issue ticket.

🦉

src/core/screen.js
AnthonyGoodchild: ⬆️

AnthonyGoodchild (approved):
This seems like a more robust solution. Some kind of lookup would be nicer, but it's only `shop` and `manage` hard coded anyway so seems suitable!

alex-s2:
There is still one reliance on checking the scene key - I've added a note to the documentation saying the shops should contain 'shop' in the key. Even if the key doesn't match you'll still get the correct stats, you'll just get some redundant outdated ones too.

Going to merge this in now as it works well enough to match the ACs.

PR #466, opened 2021-04-19
[CGPROD-3199] Fix screenshot/genie cypress tests

AdamXA:
also fixes local dev commands


AdamXA:
there is a bug in genie if you open a debug screen before a collection loads, fixing that now as part of this so I can remove the `cy.wait`

AdamXA:
> there is a bug in genie if you open a debug screen before a collection loads, fixing that now as part of this so I can remove the `cy.wait`

fixed

alex-s2 (approved):

PR #467, opened 2021-04-21
Cypress shop updates

alexparry-brown:
-re-enable cypress tests for shop
-update stats for shop to align with previous work


AnthonyGoodchild (approved):
Lgtm

PR #470, opened 2021-05-11
[CGPROD-3197] Fix button click sound while scrolling panel

AdamXA:


build-scripts/build.sh
```diff
@@ -47,7 +47,8 @@ npm run test
 if [ "$GIT_BRANCH" = "master" ]
   then npm run build:theme2
 fi
-npm run build
+npm run jest
+npm run eslint
```
AdamXA, line 51: this seems to fix eslint not failing the build when there are linting issues

build-scripts/build.sh
```diff
@@ -40,7 +40,8 @@
 #node -v
 npm install --force
 #python build-scripts/licensechecker/licensechecker.py
-npm run test
+npm run jest
+npm run eslint
```
AdamXA, line 44: this seems to fix eslint not failing the build when there are linting issues

AnthonyGoodchild (approved):
this seems like an improvement.

PR #473, opened 2021-05-14
Cgprod 3150 lock state

alex-s2:


src/components/shop/scrollable-list/scrollable-list-buttons.js
```diff
@@ -42,6 +41,8 @@ export const createListButton = (scene, item, title, action, parent) => {
     const ariaLabel = `${item.state ? item.state + " " : ""}${item.title} - ${item.subtitle}`;
     const channel = buttonsChannel(scene);
     const group = scene.scene.key;
+    const disabled = item.state && scene.config.states[item.state] ? scene.config.states[item.state].disabled : false;
+    defaults.key = disabled ? "itemBackgroundLocked" : "itemBackground";
```
AdamXA, line 45: rather than add a new asset, would this work?
`disabled && gelButton.off(Phaser.Input.Events.POINTER_OVER);`

AdamXA (commented):
you need to fix for the select screen too I think?

src/components/shop/scrollable-list/scrollable-list-buttons.js
alex-s2: its technically the same asset file just only using 1 frame of it but yea i guess that's simpler, will have a look

src/components/select/select-screen.js
```diff
@@ -76,6 +76,7 @@ export class Select extends Screen {
 
             config.suffix && (button.config.ariaLabel = [button.config.ariaLabel, config.suffix].join(" "));
             config.enabled === false && button.off(Phaser.Input.Events.POINTER_UP);
+            config.enabled === false && button.off(Phaser.Input.Events.POINTER_OVER);
```
AdamXA, line 79: haha is it called enabled in select screen and disabled in shop? 😂

AdamXA (approved):

src/components/select/select-screen.js
alex-s2: yes

PR #474, opened 2021-05-14
CGPROD-3182_Button-Hover-Bug-In-Shop

Leonie-:
- Ensures game buttons do not have hit area padding.
- Updates the buttons with larger images so they don't have to be scaled up anymore.


FostUK (approved):
Looks great! 👽

PR #475, opened 2021-05-17
CGPROD-3182_Button-Hover-Bug-In-Shop

Leonie-:
- Ensures shop uses the correct font (ReithSans) all the way through .
- Adds padding to buttons on the shop confirm screen.


Leonie-:
Reviewed/paired on with Nick 👍

PR #485, opened 2021-06-01
[CGPROD-3238] Fix distorted audio

AdamXA:
SoundManager can add multiple sounds under the same asset key (allows for different sound configs for the same asset).
So calling `scene.sound.add()` adds an extra sound to the manager each time.
We were calling `scene.sound.add()` on each button in the scene, so after loading the scene multiple times, there were a lot of sounds added, causing the distorted audio.

`.get()` will get the first sound
`.getAll()` would return an array of sounds under that key.
 https://photonstorm.github.io/phaser3-docs/Phaser.Sound.WebAudioSoundManager.html#get__anchor

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,10 @@
 
 | Version | Description |
 |---------|-------------|
-| 3.9.11 | |
+| 3.9.2 | |
+| | Updated Phaser to 3.55.2 |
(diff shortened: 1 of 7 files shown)


alex-s2:
So if it already exists it gets the existing sound with that key instead of adding it again?

Good fix, 👍

alex-s2 (approved):

AdamXA:
pretty much yea

PR #489, opened 2021-06-08
Use patched Phaser. Should move back to official builds when possible…

SmileyJames:
…. Patch required for games to work on iOS UAP apps.

Makes use of @consolenaut 's merged change:  https://github.com/photonstorm/phaser/commit/cea3d45719f0fb2b54e121c461fcea66e09833d3


AdamXA:
possible to just leave this patch in the story component? like edit the package-lock so it just resolves to this temporary npm package for that component, for now

SmileyJames:
@AdamXA that's a good idea - definetly would work for storytime. It's installed via github uri - so maybe can reference this branch.

I know we also need this patch in a couple of Hey Duggie games for Playtime Island iOS release, think they are Tier 1 games (not sure checking) so not using genie.

SmileyJames:
Prefer: https://github.com/bbc/genie/pull/489

PR #493, opened 2021-06-25
Sets isKindle when inside app web view on Kindle

FostUK:
Sets isKindle when inside app web view on Kindle


src/core/browser.js
```diff
@@ -15,7 +15,10 @@ export const getBrowser = () => {
 
     const name = browserInfo.getBrowserName();
     const version = browserInfo.getBrowserVersion();
-    const isSilk = name === "Amazon Silk";
+    const isKindleWebView = new RegExp(
+        /KFOT|KFTT|KFJWI|KFJWA|KFOTE|KFSOWI|KFTHWI|KFTHWA|KFAPWI|KFAPWA|WFJWAE|KFSAWA|KFSAWI|KFASWI|KFARWI|KFFOWI|KFGIWI|KFMEWI/g,
```
AnthonyGoodchild, line 19: 🤔

AnthonyGoodchild: ah android build refs or something?

AnthonyGoodchild (approved):
looks good just a linting fix 👍

src/core/browser.js
FostUK: Yeah list supplied by apps. Not sure any other ways to narrow it down exist. Was hoping could add " Build" to the end but that may or may not be present.

PR #494, opened 2021-06-29
Fix isKindleWebView from returning different boolean each time

AdamXA:
Reusing the same RegExp object was causing it to return a different boolean each time. Couldn't reproduce this behaviour in unit tests though, bit strange.


FostUK (approved):
Looks great!

PR #500, opened 2021-11-04
Update packages and linting

FostUK:
* Update packages to clear security alerts
* New linting config for latest eslint


babel.config.js
```diff
@@ -35,6 +35,7 @@ module.exports = function (api) {
 				},
 			},
 		],
+		["@babel/plugin-proposal-private-property-in-object", { loose: false }],
```
AnthonyGoodchild, line 38: try saying that with a mouth full of crisps

FostUK: Tbf - I'm not entirely sure what's going on with that other than it spams errors if youi don't have it.

AnthonyGoodchild (approved):
Nice one

PR #519, opened 2022-11-03
CGPROD-3752 added config option for default home screen

sophia4422:
Allows to configure different default screen from the home.


alex-s2 (approved):
👍

PR #529, opened 2023-06-29
[CGPROD-3889] level select

juditlehoczki:
Ticket: https://jira.dev.bbc.co.uk/browse/CGPROD-3889

### DIFF
--- a/cypress/integration/select_spec.js
+++ b/cypress/integration/select_spec.js
@@ -89,4 +89,43 @@ describe(`The Select screens ${Cypress.env("THEME")}`, () => {
 		cy.ariaHidden("#debug-select-grid__mary", "true");
 		cy.ariaHidden("#debug-select-grid__mike", "false");
 	});
+
+	it("Scrolls to the page where the last selected item is within the same session", () => {
+		cy.genieClick("#home__debug");
+		cy.genieClick("#debug__debug-select-grid");
+		cy.genieClick("#debug-select-grid__next").wait(2000);
+		cy.genieClick("#debug-select-grid__mike").wait(2000);
+		cy.genieClick("#debug__debug-select-grid").wait(2000);
+		cy.ariaHidden("#debug-select-grid__mary", "true");
+		cy.ariaHidden("#debug-select-grid__kyle", "true");
+		cy.ariaHidden("#debug-select-grid__suzy", "true");
+		cy.ariaHidden("#debug-select-grid__greg", "true");
+		cy.ariaHidden("#debug-select-grid__ezinma", "true");
+		cy.ariaHidden("#debug-select-grid__phil", "true");
+		cy.ariaHidden("#debug-select-grid__mike", "false");
+		cy.ariaHidden("#debug-select-grid__debby", "false");
+		cy.ariaHidden("#debug-select-grid__timmy", "false");
+		cy.ariaHidden("#debug-select-grid__elliot", "false");
+	});
+
+	it("Scrolls to the page where the last selected item is after a page-refresh", () => {
+		cy.genieClick("#home__debug");
+		cy.genieClick("#debug__debug-select-grid");
+		cy.genieClick("#debug-select-grid__next").wait(2000);
+		cy.genieClick("#debug-select-grid__mike").wait(2000);
+		cy.visit(appendToken(`${Cypress.env("url")}${getUrl()}`));
+		cy.get("#home__play", { timeout: 60000 }).should("exist");
+		cy.genieClick("#home__debug");
+		cy.genieClick("#debug__debug-select-grid").wait(2000);
+		cy.ariaHidden("#debug-select-grid__mary", "true");
+		cy.ariaHidden("#debug-select-grid__kyle", "true");
+		cy.ariaHidden("#debug-select-grid__suzy", "true");
+		cy.ariaHidden("#debug-select-grid__greg", "true");
+		cy.ariaHidden("#debug-select-grid__ezinma", "true");
+		cy.ariaHidden("#debug-select-grid__phil", "true");
+		cy.ariaHidden("#debug-select-grid__mike", "false");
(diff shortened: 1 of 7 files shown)


src/core/collections.js
```diff
@@ -59,11 +59,27 @@ export const initCollection = screen => key => {
 		gmi.setGameData("genie", fp.setWith(Object, storagePath, mergeToList(getStored(), [config]), getGenieStore()));
 	};
 
+	const setUnique = ({ id, key, value }) => {
```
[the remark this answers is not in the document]
juditlehoczki: Yeah, it was in my mind to change `doesItNeedToBeReset` but then I totally forgot.. 😅 I already went with Alexa's suggestion of `shouldBeReset`. 👍

src/core/collections.js
```diff
@@ -59,11 +59,27 @@ export const initCollection = screen => key => {
...
+			const isNewlyUnique = item.id === id && item[key] !== value;
+			const shouldBeReset = item.id !== id && item.hasOwnProperty(key) && item[key] === value;
+
+			if (isNewlyUnique) {
+				set({ id, [key]: value });
+				item[key] = value;
+			} else if (shouldBeReset) {
+				set({ id: item.id, [key]: null });
+				item[key] = null;
+			}
```
AnthonyGoodchild, line 73: would this work if it was simplified to just update the value if

```
item.id === id && item[key]
```

and set to null if not?

AnthonyGoodchild: not suggesting you do this, but the following seems to work, with the caveat of adding 'null' to all non-selected items:

```suggestion
			const isNewlyUnique = item.id === id;

			if (isNewlyUnique) {
				set({ id, [key]: value });
				item[key] = value;
			} else {
				set({ id: item.id, [key]: null });
				item[key] = null;
			}
```

I guess that was what you were trying to avoid?

AnthonyGoodchild: it's a shame we can't unset the value - it would be much cleaner in the saveData object if we could set undefined.

test/components/select/select-screen.test.js
```diff
@@ -337,6 +338,13 @@ describe("Select Screen", () => {
 			eventBus.subscribe.mock.calls[0][0].callback();
 			expect(selectScreen.transientData["test-select"].choice.title).toBe("Title 1");
 		});
+		test("saves choice to local storage", () => {
```
AnthonyGoodchild, line 341: might be better name "sets choice to collection" - local storage is kind of abstract here

cypress/integration/select_spec.js
```diff
@@ -89,4 +89,43 @@ describe(`The Select screens ${Cypress.env("THEME")}`, () => {
...
+		cy.ariaHidden("#debug-select-grid__kyle", "true");
+		cy.ariaHidden("#debug-select-grid__suzy", "true");
+		cy.ariaHidden("#debug-select-grid__greg", "true");
+		cy.ariaHidden("#debug-select-grid__ezinma", "true");
+		cy.ariaHidden("#debug-select-grid__phil", "true");
+		cy.ariaHidden("#debug-select-grid__mike", "false");
+		cy.ariaHidden("#debug-select-grid__debby", "false");
+		cy.ariaHidden("#debug-select-grid__timmy", "false");
+		cy.ariaHidden("#debug-select-grid__elliot", "false");
+	});
```
AnthonyGoodchild, line 130: We should do more of this, there are probably a lot of things we aren't testing from a behaviour/journey perspective

src/core/collections.js
juditlehoczki: We considered this with Nick but then decided against it so we don't append an unnecessary `[key]: null` to each element that doesn't need it.

test/components/select/select-screen.test.js
juditlehoczki: I thought it would be better to be more specific here because you could update 'collection' but not the 'local storage' and viceversa?

juditlehoczki: actually, the point here is to set both :D

AnthonyGoodchild: it's not the concern of the select screen how the data is handled though - that part of the implementation is down to `collection`.  The fact that you are mocking collection highlights this too - if the actual implementation of collection changed then this test would still pass, even if collection started using cookies or a database for example.

AnthonyGoodchild: sorry collection doesn't even care how it's saved either - that's also just a GMI call, so it's only the GMI that handles the localstorage bit.

Obviously we know it's using the local storage, but compartmentalising the units and making everything else a black box is good for unit testing!

juditlehoczki: I've changed it now

AnthonyGoodchild: > the point here is to set both

the method and tests for `setUnique` handles that, so you can assume it's doing that here.

AnthonyGoodchild: sorry for spam, just trying to explain myself best I can 😄

test/core/collections.test.js
```diff
@@ -308,4 +308,30 @@ describe("Collections", () => {
...
+	describe("Returned setUnique method", () => {
+		test("Sets value on one item in collection", () => {
+			const collection = initCollection(mockScreen)("testCollection");
+			collection.setUnique({ id: "id1", key: "selected", value: true });
+
+			expect(mockGmi.setGameData.mock.calls[0][1].collections.testCollection[0]).toEqual({
+				id: "id1",
+				selected: true,
+			});
+			expect(collection.getAll()[0].selected).toBe(true);
```
AnthonyGoodchild, line 321: nitpick but you could assert that the other item doesn't have 'selected' on it, for added robustness

AnthonyGoodchild (approved):
LGTM 👍

PR #532, opened 2023-07-28
[CGPROD-3889] remove transient data for selections

juditlehoczki:
- remove transient data for selections stored in local storage
- update failing tests
- add test for `getUnique`

### DIFF
--- a/docs/development/select-screen.md
+++ b/docs/development/select-screen.md
@@ -120,10 +120,10 @@ The select screen will paginate the choices based on the `rows` and `cols` param
 }
 ```
 
-The page displayed when first loaded can be changed by setting a unique key on the element in the collection, e.g:
+The page displayed when first loaded can be changed by setting a `selected` key to `true` on the element in the collection, e.g:
 
 ```javascript
-this.collection.setUnique({ id: "1", key: "default", value: true });
+this.collection.setUnique({ id: "1", key: "selected", value: true });
 ```
 
 The select screen will then display the appropriate page for this choice when navigated.
--- a/docs/development/select-screen.md
+++ b/docs/development/select-screen.md
@@ -94,7 +94,7 @@ Configured choices may have text labels drawn with them. These are populated by 
 
 ### Choice states
 
-Choices can have states, see [states](docs/development/collections.md). 
+Choices can have states, see [states](collections.md). 
 
 ```json5
 {
@@ -120,10 +120,10 @@ The select screen will paginate the choices based on the `rows` and `cols` param
 }
 ```
 
-The page displayed when first loaded can be changed by setting `choice` in transient data, under the key for the screen, e.g:
+The page displayed when first loaded can be changed by setting a unique key on the element in the collection, e.g:
 
 ```javascript
-this.transientData["level-select"].choice = { id: "dennis" }
+this.collection.setUnique({ id: "1", key: "default", value: true });
 ```
 
 The select screen will then display the appropriate page for this choice when navigated.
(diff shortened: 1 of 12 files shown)


src/core/collections.js
```diff
@@ -74,12 +74,18 @@ export const initCollection = screen => key => {
 		});
 	};
 
+	const getUnique = ({ key, value }) => catalogue.find(item => item[key] === value);
+
+	const getSelected = () => getUnique({ key: "selected", value: true });
```
FostUK, line 79: This to be more generic.

Collections are (supposed to be) agnostic so having a selection specific method for a tag starts adding hard dependencies between the select screen's domain and the collection domain (I am aware the `qty` value breaks this idea though :) )

AnthonyGoodchild (commented):
I think we can get rid of this test now:
`saves choice to transient data`

alex-s2 (approved):

test/game.test.js
```diff
@@ -158,17 +155,31 @@ describe("Game", () => {
...
 				align: "center",
 			});
 		});
 
 		test("displays the level selected", () => {
-			const expectedLevel = `Level Selected: ${game.transientData["level-select"].choice.title}`;
-			expect(game.add.text).toHaveBeenCalledWith(150, 250, expectedLevel, {
+			const levelTitle = "Hard level";
+			const expected = `Level Selected: ${levelTitle}`;
+			let mockGetUnique = jest.fn().mockImplementation(() => ({ title: levelTitle }));
```
FostUK, line 176: I think mockImplenentation can just be replaced these days by passing it into the functione.g:

`jest.fn(() => ({ title: levelTitle }));`

FostUK (approved):
Looks great - added a minor note about mockImplementation you could just use in the future - maybe not worth changing this.

test/components/select/select-screen.test.js
```diff
@@ -331,13 +332,6 @@ describe("Select Screen", () => {
 			jest.spyOn(eventBus, "subscribe");
 		});
 
-		test("saves choice to transient data", () => {
-			mockChoices = [{ title: "Title 1" }];
-			selectScreen.create();
-
-			eventBus.subscribe.mock.calls[0][0].callback();
-			expect(selectScreen.transientData["test-select"].choice.title).toBe("Title 1");
-		});
```
AnthonyGoodchild, line 340: I think this might still be a valid test. Sorry, my fault 😂 

We do utilise 'selected' transientData for referencing the confirmed selection, i.e so the game knows what was selected:

https://github.com/bbc/genie/pull/532/files#diff-38a518a8d1ac1c089e299b6b631a4fe07a57a218dc99e27970da8800d1aa5d0bR102

juditlehoczki: I've put it back 😄

AnthonyGoodchild (approved):

docs/development/select-screen.md
```diff
@@ -120,10 +120,10 @@ The select screen will paginate the choices based on the `rows` and `cols` param
 }
 ```
 
-The page displayed when first loaded can be changed by setting `choice` in transient data, under the key for the screen, e.g:
+The page displayed when first loaded can be changed by setting a unique key on the element in the collection, e.g:
```
AnthonyGoodchild, line 123: Doesn't the value need to be 'selected' for the select screen to display it?

AnthonyGoodchild: not value, key, e.g: 

```{ id: "1", key: "selected", value: true }```

juditlehoczki: yes... I'm having a moment lol

alex-s2 (approved):

AnthonyGoodchild (approved):
LGTM 👍

AnthonyGoodchild (changes requested):
Could do with bumping the minor version, as this is a breaking change to the collection/select screen API

PR #535, opened 2023-08-15
[CGPROD-3891] top right buttons - vertical option

juditlehoczki:
Ticket: https://jira.dev.bbc.co.uk/browse/CGPROD-3891

- added global `game-config` file with a flag for `verticalTopRight` that loads in in `Boot`
- added `topRightV` group layout option
- fixed offset/positioning of top right buttons when vertical layout is selected
- updated/added relating unit tests

### DIFF
--- a/src/core/layout/gel-defaults.js
+++ b/src/core/layout/gel-defaults.js
@@ -20,7 +20,7 @@ const addChannelToAll = (defaults, channel) => Object.fromEntries(Object.entries
 
 export const buttonsChannel = screen => (screen ? `gel-buttons-${screen.scene.key}` : "gel-buttons");
 export const config = screen => {
-	const { verticalTopRight } = screen?.cache.json.get("game-config") || {};
+	const { topRightVertical } = screen?.cache.json.get("config") || {};
 
 	const gelDefaults = {
 		exit: {
@@ -72,7 +72,7 @@ export const config = screen => {
 			},
 		},
 		audio: {
-			group: verticalTopRight ? "topRightV" : "topRight",
+			group: topRightVertical ? "topRightV" : "topRight",
 			title: "Sound Off",
 			key: "audio-on",
 			ariaLabel: "Toggle Sound",
@@ -91,7 +91,7 @@ export const config = screen => {
 			},
 		},
 		settings: {
-			group: verticalTopRight ? "topRightV" : "topRight",
+			group: topRightVertical ? "topRightV" : "topRight",
 			title: "Settings",
 			key: "settings",
 			ariaLabel: "Game Settings",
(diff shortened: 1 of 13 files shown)


test/core/layout/layout.test.js
```diff
@@ -211,6 +213,29 @@ describe("Layout", () => {
...
+
+		test("adds settings icons in the topRightV group layout when verticalTopRight config is set true", () => {
+			mockScene.config.subtitle = undefined;
+			mockJson.verticalTopRight = true;
+			const layout = Layout.create(mockScene, mockMetrics, ["settings"]);
+
+			expect(layout.buttons.settings.buttonName.group).toBe("topRightV");
+		});
+
+		test("adds settings icons in the topRightV group layout when verticalTopRight config is not provided", () => {
```
AnthonyGoodchild, line 231: should this be `topRight group` ?

test/core/layout/layout.test.js
```diff
@@ -15,6 +15,7 @@ jest.mock("../../../src/core/layout/gel-group.js");
 
 describe("Layout", () => {
 	const sixGelButtons = ["achievements", "exit", "howToPlay", "play", "audio", "settings"];
+	const expectedNumberOfGroupLayouts = 12;
```
FostUK, line 18: `expectedLayouts`

Added a note that we should discuss naming variables at the next dev meet - our approach is to try and be concise where possible.

Also could you do me a favour and rename `sixGelButtons` to `gelButtons` as that's an awful name that it appears I added 😂

test/core/layout/layout.test.js
AnthonyGoodchild:
```suggestion
		test("adds settings icons in the topRight group layout when verticalTopRight config is not provided", () => {
```

src/core/layout/gel-defaults.js
```diff
@@ -20,6 +20,8 @@ const addChannelToAll = (defaults, channel) => Object.fromEntries(Object.entries
 
 export const buttonsChannel = screen => (screen ? `gel-buttons-${screen.scene.key}` : "gel-buttons");
 export const config = screen => {
+	const { verticalTopRight } = screen?.cache.json.get("game-config") || {};
```
FostUK, line 23: As this is not just for the game component, but is more global, could we call this `config.json` ? It's already at top level in the theme folder so this should make sense.

src/core/layout/gel-defaults.js
```diff
@@ -70,7 +72,7 @@ export const config = screen => {
 			},
 		},
 		audio: {
-			group: "topRight",
+			group: verticalTopRight ? "topRightV" : "topRight",
```
FostUK, line 75: Yeah maybe just match the order so `topRightVertical` for the config value.

_(This is probably the most pedantic comment I have ever left)_

FostUK (approved):
This looks great. Well done on getting your head round the layout system - realise it's quite a complex beast!

AnthonyGoodchild (approved):
Looks good to me ⬆️

FostUK (approved):
Looks great! 🍤 🍚🍖🍥🏮

PR #537, opened 2023-08-18
CGPROD-3888 Add Voice Over to Gel Buttons

sophia4422:
For spike ticket: [CGPROD-3888](https://jira.dev.bbc.co.uk/browse/CGPROD-3888)

I have added temporary mp4s for now which make a sound when you hover over the button. 
(Not all of the buttons have sounds, as waiting on proper audio - sound has been taken from another game)


themes/default/character-select/config.json5
AnthonyGoodchild, line 1: I'm not 100% sure as on mobile but has your editor reformatted all the existing config? The diff might be a lot smaller if it wasn't.

AnthonyGoodchild:
Looks good, it would be nice to see if you could sort the diff out though on all the config changes

PR #543, opened 2023-10-16
10P Custom Resolution

FostUK:
Add method to multiply resolution.

Potentially needs launching via config in the future.


AnthonyGoodchild (approved):
👍

PR #548, opened 2023-10-20
CGPROD 3891 Slow local dev startup

FostUK:
Disable compression (was causing slow starts for local dev when using unbundled ES6 modules)


AnthonyGoodchild (approved):
⚡ ⚡

PR #556, opened 2023-12-14
[CGPROD-4040] call ShowCurrentPage after adding screens

juditlehoczki:
https://jira.dev.bbc.co.uk/browse/CGPROD-4040


FostUK (commented):
Good find but needs a test :)

FostUK (approved):
Looks great!
🦆

